PR #1, opened 2016-08-09
Comments

helfer:
Making a PR so I can add comments to the code.


index.js
```diff
@@ -0,0 +1,6 @@
+var Client = require('./src/client.js');
```
helfer: const Client = import './src/client.js';

export { Client };

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+
+    this.client.onclose = function() {
+      console.log("connection closed");
+    };
+
+    this.client.onerror = function(error) {
+      console.log('connection error: ', error);
+    };
+  }
+  openConnection(handler) {
```
helfer: remove this

src/client.js
```diff
@@ -0,0 +1,97 @@
+"use strict";
+var W3CWebSocket = require('websocket').w3cwebsocket;
+class Client {
+  constructor(url, protocol) {
+    this.url = url;
```
helfer: don't need to store this.url and this.protocol

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+    this.protocol = protocol;
+    this.client = new W3CWebSocket(url, protocol);
+    this.subscriptions = {}; // id: handler
+    this.max_id = 0;
+
+    this.client.onopen = () => {
+      console.log("Websocket client connected.");
```
helfer: get rid of logging for production

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    this.client.onmessage = (message) => {
+      let message_data = JSON.parse(message.data);
```
helfer: use CamelCase

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+      if (message_data.type === 'subscription_fail') {
```
helfer: use constants for 'subscription_fail', 'subscription_data', etc, and import them in server and client.

helfer: turn it into a switch statement maybe

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+        const del_id = message_data.id;
+        delete this.subscriptions[del_id];
```
helfer: call handler with error.

src/client.js
helfer: do we need try/catch for JSON.parse?

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+      } else if (message_data.type === 'subscription_data') {
+        if (message_data.data) {
+          const sub_id = message_data.id;
+          if (this.subscriptions[sub_id]) {
+            this.subscriptions[sub_id](null, message_data.data); // pass data into data handler
```
helfer: maybe call this.subscriptions this.subscriptionHandlers if only handlers are stored.

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+          }
+        } else {
+          this.subscriptions[sub_id](message_data.errors, null);
+        }
+      }
+    };
+
+    this.client.onclose = function() {
+      console.log("connection closed");
```
helfer: Should we clean up subscription data?

helfer: Should there be a way for the server to stop a client subscription, for example with a 'subscription_stop' message?

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+    if (!handler) {
+      handler = (error) => {console.log('Error opening connection', error);};
+    }
+    this.client.on('connectFailed', function(error) {
+      handler(error);
+    });
+    this.client.connect(this.url, this.protocol);
+  }
+
+  sendMessage(message) {
```
helfer: maybe call it _sendMessage

src/client.js
```diff
@@ -0,0 +1,97 @@
+"use strict";
+var W3CWebSocket = require('websocket').w3cwebsocket;
+class Client {
+  constructor(url, protocol) {
+    this.url = url;
+    this.protocol = protocol;
+    this.client = new W3CWebSocket(url, protocol);
+    this.subscriptions = {}; // id: handler
+    this.max_id = 0;
```
helfer: Are these private? If so, then call them _client etc.

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+  options:
+    - query
+    - variables
+    - operationName
+    - pollingInterval
+    - triggers
+  */
+  subscribe(options, handler) {
+    if (this.client.readyState !== this.client.OPEN) {
+      throw new Error('Client is not connected to a websocket.');
```
helfer: Maybe be more specific about what the readyState is, that way the caller can handle these different cases differently.

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+
+  /*
+  options:
+    - query
+    - variables
+    - operationName
+    - pollingInterval
+    - triggers
+  */
+  subscribe(options, handler) {
```
helfer: Check arguments here, especially options, and throw an error if required arguments are missing or the wrong type.

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+      message.id = sub_id;
+      this.sendMessage(message);
+      this.subscriptions[sub_id] = handler;
+      return sub_id;
+    }
+  }
+
+  unsubscribe(id) {
+    if (this.client.readyState !== this.client.OPEN) {
+      throw new Error('Client is not connected to a websocket.');
```
helfer: Same thing with readyState

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+      let message_data = JSON.parse(message.data);
+      if (message_data.type === 'subscription_fail') {
+        const del_id = message_data.id;
+        delete this.subscriptions[del_id];
+      } else if (message_data.type === 'subscription_data') {
+        if (message_data.data) {
+          const sub_id = message_data.id;
+          if (this.subscriptions[sub_id]) {
+            this.subscriptions[sub_id](null, message_data.data); // pass data into data handler
+          }
```
helfer: If this.subscriptions[sub_id] doesn't exist, maybe we should send a 'subscription_end' to the server.

src/client.js
helfer: Or maybe we can just delete locally without sending the message. If server sends data later, we'll just reply with a 'subscription_end'.

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  unsubscribe(id) {
+    if (this.client.readyState !== this.client.OPEN) {
+      throw new Error('Client is not connected to a websocket.');
+    } else {
+      let message = { id: id, type: 'subscription_end'};
+      this.sendMessage(message);
+      if (this.subscriptions[id]) {
```
helfer: don't need the 'if' here.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+  options {
+    schema: GraphQLSchema
+    contextValue?: any,
+    rootValue?: any,
+    formatResponse?: (Object) => Object,
+    validationRules?: Array<any> 
+    triggerGenerator?: (Object) => [Object]
+  }
+  */
+  constructor(options, httpServer) {
```
helfer: Check arguments and provide useful error messages.

src/server.js
```diff
@@ -0,0 +1,150 @@
+"use strict";
+var http = require('http');
+var graphql = require('graphql');
+require('babel-register');
+var WebSocketServer = require('websocket').server;
+var graphql_tools = require('graphql-tools');
+var graphql_validator = require('graphql/validation');
+var graphql_execution = require('graphql/execution');
```
helfer: I think you can just import validate and execute directly.
`import { execute } from 'graphql';`

test/client.js
```diff
@@ -0,0 +1,539 @@
+var assert = require('chai').assert;
+var Client = require('../src/client.js');
+var graphql_tools = require('graphql-tools');
```
helfer: unused import

src/server.js
```diff
@@ -0,0 +1,150 @@
+"use strict";
+var http = require('http');
+var graphql = require('graphql');
+require('babel-register');
+var WebSocketServer = require('websocket').server;
+var graphql_tools = require('graphql-tools');
```
helfer: unused import

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+var graphql_validator = require('graphql/validation');
+var graphql_execution = require('graphql/execution');
+
+class Server {
+  /*
+  options {
+    schema: GraphQLSchema
+    contextValue?: any,
+    rootValue?: any,
```
helfer: We'll need to provide contextValue and rootValue per connection. Probably also some kind of functionality for logging in and logging out.

src/client.js
```diff
@@ -0,0 +1,97 @@
+"use strict";
+var W3CWebSocket = require('websocket').w3cwebsocket;
+class Client {
+  constructor(url, protocol) {
```
helfer: protocol doesn't need to be an option, because it only changes if the code changes.
Maybe call it 'graphql-subscription' or 'graphql-subscription-protocol'

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+
+    //init and connect websocket server to http
+    this.wsServer = new WebSocketServer({
+      httpServer: httpServer,
+      autoAcceptConnections: false
+    });
+
+    this.wsServer.on('request', (request) => {
+      //accept connection
+      var connection = request.accept('graphql-protocol', request.origin);
```
helfer: Maybe provide a way for users to restrict origin?
And probably should only accept connection if it's asking for 'graphql-protocol'.
Not sure, but maybe protocol will also have a verison in the future.
Think about whether protocol name should be an option to the server?

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+            let message = {
+              type: 'subscription_fail',
+              errors: syntax_errors,
+              id: message_data.id,
+            };
+            connection.sendUTF(JSON.stringify(message));
+          } else {
+            const sub_id = message_data.id;
+            if (!connection.subscriptions) {
+              connection.subscriptions = {};
```
helfer: maybe not store stuff on the connection object? I would store it on the Server in this.subscriptions[connection] or something. If it's not possible to get a short identifier for the connection, then maybe store it on connection after all...

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+    this.wsServer = new WebSocketServer({
+      httpServer: httpServer,
+      autoAcceptConnections: false
+    });
+
+    this.wsServer.on('request', (request) => {
+      //accept connection
+      var connection = request.accept('graphql-protocol', request.origin);
+      console.log('Accepted connection');
+      connection.on('message', (message) => {
```
helfer: I would break this stuff here out into a separate function so it's easier to read.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+        let message_data = JSON.parse(message.utf8Data);
+        if (message_data.type === 'subscription_start') {
```
helfer: Do a switch statement and then maybe a single function inside each case to handle that type of message.

src/client.js
```diff
@@ -0,0 +1,97 @@
...
+    this.max_id += 1;
+    return id;
+  }
+
+  /*
+  options:
+    - query
+    - variables
+    - operationName
+    - pollingInterval
```
helfer: Let's get rid of polling interval

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+            }
+            msg_triggers.forEach((trigger) => {
+              let string_trigger = JSON.stringify(trigger);
+              if (! this.triggers[string_trigger]) {
+                this.triggers[string_trigger] = [{connection: connection, sub_id: sub_id}];
+              } else {
+                this.triggers[string_trigger].push({connection: connection, sub_id: sub_id});
+              }
+            });
+            //set up polling message
```
helfer: We'll get rid of this. Triggers are enough.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+            }
+            //set up trigger listeners
+            let msg_triggers = [];
+            if (message_data.triggers) {
+              msg_triggers = message_data.triggers;
+
+            } else {
+              if (this.options.triggerGenerator) {
+                msg_triggers = this.options.triggerGenerator(message_data);
+              }
```
helfer: If triggers are not provided and there's no trigger generator, then I think the subscription should fail.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+
+class Server {
+  /*
+  options {
+    schema: GraphQLSchema
+    contextValue?: any,
+    rootValue?: any,
+    formatResponse?: (Object) => Object,
+    validationRules?: Array<any> 
+    triggerGenerator?: (Object) => [Object]
```
helfer: We probably need to document this.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+    });
+
+    this.wsServer.on('request', (request) => {
+      //accept connection
+      var connection = request.accept('graphql-protocol', request.origin);
+      console.log('Accepted connection');
+      connection.on('message', (message) => {
+        let message_data = JSON.parse(message.utf8Data);
+        if (message_data.type === 'subscription_start') {
+          let syntax_errors = graphql_validator.validate(this.options.schema, graphql.parse(message_data.query), this.options.validationRules);
```
helfer: make sure that validationRules gets concatenated with the default graphql validation rules before calling validate.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+      httpServer: httpServer,
+      autoAcceptConnections: false
+    });
+
+    this.wsServer.on('request', (request) => {
+      //accept connection
+      var connection = request.accept('graphql-protocol', request.origin);
+      console.log('Accepted connection');
+      connection.on('message', (message) => {
+        let message_data = JSON.parse(message.utf8Data);
```
helfer: This could throw an error, we should catch it and then return a subscription_fail.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+            if (message_data.triggers) {
+              msg_triggers = message_data.triggers;
+
+            } else {
+              if (this.options.triggerGenerator) {
+                msg_triggers = this.options.triggerGenerator(message_data);
+              }
+            }
+            msg_triggers.forEach((trigger) => {
+              let string_trigger = JSON.stringify(trigger);
```
helfer: Make sure oder of object properties is always the same when serialized.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+          
+        } else if (message_data.type === 'subscription_end') {
+          const sub_id = message.data.id;
+          clearInterval(connection.subscriptions[sub_id][pollingId]);
+          delete connection.subscriptions[sub_id];
+        }
+      });
+
+      connection.on('close', function(reasonCode, description) {
+        console.log(" Peer " + connection.remoteAddress + ' disconnected.');
```
helfer: should probably clean subscriptions here.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+              message_data.pollingId = pollingId;
+            }
+
+            connection.subscriptions[sub_id] = message_data;
+          }
+          
+        } else if (message_data.type === 'subscription_end') {
+          const sub_id = message.data.id;
+          clearInterval(connection.subscriptions[sub_id][pollingId]);
+          delete connection.subscriptions[sub_id];
```
helfer: have to clean up triggers at some point when client unsubscribes.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+
+            } else {
+              if (this.options.triggerGenerator) {
+                msg_triggers = this.options.triggerGenerator(message_data);
+              }
+            }
+            msg_triggers.forEach((trigger) => {
+              let string_trigger = JSON.stringify(trigger);
+              if (! this.triggers[string_trigger]) {
+                this.triggers[string_trigger] = [{connection: connection, sub_id: sub_id}];
```
helfer: for cleaning up it would be easier to store this.triggers[stringTrigger][connectionId]. That's hoping that there's something like a connection id.

src/server.js
helfer: Generate an id for each subscription, then store them in a map.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+        }
+      });
+
+      connection.on('close', function(reasonCode, description) {
+        console.log(" Peer " + connection.remoteAddress + ' disconnected.');
+      });
+    });
+  }
+
+  triggerAction(message_data) {
```
helfer: I think messageData is a bit confusing as argument name here. maybe triggerObject if it's the same as before?

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+    if (this.triggers[JSON.stringify(message_data)]) {
```
helfer: same thing about serialization here.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+      let triggered_subs = this.triggers[JSON.stringify(message_data)];
+      triggered_subs.forEach((sub_obj) => {
+        let sub_connection = sub_obj.connection;
```
helfer: const { connection, subId } = subscriptionObject;

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+        let sub_id = sub_obj.sub_id;
+        let sub_data = sub_connection.subscriptions[sub_id];
```
helfer: call this subscriptionArguments, it will be easier to understand, I think.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+            });
+            //set up polling message
+            if (message_data.pollingInterval) {
+              let pollingId = setInterval(
+                () => {
+                  graphql.graphql(
+                    this.options.schema,
+                    message_data.query,
+                    this.options.rootValue,
+                    this.options.contextValue,
```
helfer: we'll probably store root and context per connection, so we'll need to get that from there.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+      let triggered_subs = this.triggers[JSON.stringify(message_data)];
+      triggered_subs.forEach((sub_obj) => {
+        let sub_connection = sub_obj.connection;
+        let sub_id = sub_obj.sub_id;
+        let sub_data = sub_connection.subscriptions[sub_id];
+        graphql.graphql(
+          this.options.schema,
+          sub_data.query,
+          this.options.rootValue,
+          this.options.contextValue,
```
helfer: root and context will probably come from connection.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  triggerAction(message_data) {
+    if (this.triggers[JSON.stringify(message_data)]) {
+      let triggered_subs = this.triggers[JSON.stringify(message_data)];
+      triggered_subs.forEach((sub_obj) => {
+        let sub_connection = sub_obj.connection;
+        let sub_id = sub_obj.sub_id;
+        let sub_data = sub_connection.subscriptions[sub_id];
+        graphql.graphql(
```
helfer: moving this to a separate function will make it easier to read, I think.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+          this.options.schema,
+          sub_data.query,
+          this.options.rootValue,
+          this.options.contextValue,
+          sub_data.variables,
+          sub_data.operationName
+        ).then((response) => {
+          let message = response;
+          message.type = 'subscription_data';
+          message.id = sub_id;
```
helfer: For message, we could use this format: { type, id, payload } so type and id don't interfere with payload.

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+          if (this.options.formatResponse) {
+            message = formatResponse(message);
+          }
+          sub_connection.sendUTF(JSON.stringify(message));
+        }, (err) => {
```
helfer: you can do this with .catch (I think it's more idiomatic)

src/server.js
```diff
@@ -0,0 +1,150 @@
...
+  */
+  constructor(options, httpServer) {
+    this.options = options;
+    this.triggers = {}; // trigger_object: [{connection, sub_id}]
+    //initialize http server
+
+    //init and connect websocket server to http
+    this.wsServer = new WebSocketServer({
+      httpServer: httpServer,
+      autoAcceptConnections: false
```
helfer: look at other options to websocket server, like ignoreXForwardedFor. Do we need them?

src/server.js
helfer: Should we run the query initially anyway? Or not?

src/server.js
helfer: Also add a way for people to run a function on every subscription so they can filter query and variables etc, or restrict the kinds of subscription triggers people can set.

src/client.js
amandajliu: made it private

src/client.js
amandajliu: Did both

src/server.js
amandajliu: now stored by trigger name

src/server.js
amandajliu: seems like there isn't :/

src/client.js
amandajliu: handled by ts

helfer: That's only the case if you use it from a typescript project. If you use it from JavaScript you still don't get type safety.

src/server.js
amandajliu: handled by ts

src/server.js
amandajliu: I'm now just storing all necessary information in server.triggers (inc the connection object), and not storing anything on the connection

src/server.js
amandajliu: so instead of `triggerObject = {name: ..., value: ...}` should I have `triggerObject = {name: ..., contextValue: ..., rootValue: ...}`?

src/server.js
amandajliu: cleaned up

src/server.js
amandajliu: could you expand on this? not sure exactly what you mean - would this be a function passed into the server constructor, and then run every time a subscription_start message is sent to the server?

helfer: Yeah, either run on every subscription_start, or even better a function run on every message, kind of like middleware. But this is not super-important, so we can also add it later.

PR #3, opened 2016-09-06
add CI and coverage

helfer:


stubailo:
Don't forget to turn off comments from coveralls lol

helfer:
I really don't know why that's not the default 😕

src/server.ts
```diff
@@ -107,7 +107,7 @@ class Server {
             formatError: undefined,
             callback: undefined,
           };
-          params = this.onSubscribe(parsedMessage, params);
+          params = this.onSubscribe(parsedMessage, params) || params;
```
stubailo: IMO this code is not very readable, it would be better as an if statement with an intermediate variable.

PR #5, opened 2016-09-09
Updated graphql.js dependency to 0.7.0 so there will be no duplicates

davidyaha:
Following on this PR https://github.com/apollostack/graphql-subscriptions/pull/4

Also:
* Updated typescript to version ^2.0.0 to allow for trailing commas on function parameters (wasn't allowed on older versions)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "es6-promise": "^3.2.1",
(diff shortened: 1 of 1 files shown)


helfer:
graphql-js isn't used any more in this package, so I will remove the dependency.

PR #10, opened 2016-09-14
Fix typo in client-side SUBSCRIPTION_FAIL handling

lfittl:
Without this fix, if there are errors on the server side, the client will still run the `next(..)` handler on the client, instead of the `error(..)` handler.

This seems like a simple oversight, since the next block uses `payload.errors` correctly.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -72,7 +72,7 @@ export default class Client {
           break;
         case SUBSCRIPTION_FAIL:
           if (this.subscriptionHandlers[subId]) {
-            this.subscriptionHandlers[subId](parsedMessage.errors, null);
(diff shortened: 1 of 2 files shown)


helfer:
Thanks, good catch!

helfer:
@lfittl Do you think you could add a test for this as well? I would be eternally grateful! 😃

lfittl:
@helfer Sure, I'll give that a try later tonight/tomorrow :)

helfer:
Thanks a lot! You can push the test to this PR as well, then I'll merge it all at once.

stubailo:
Ping - let's get a test going!

lfittl:
@helfer @stubailo Test added, sorry for the delay :)

stubailo:
Awesome, thanks!

PR #11, opened 2016-09-14
fix SUBSCRIPTION_FAIL payload handling in Client

bodia-uz:
Pass `parsedMessage.payload.errors` instead undefined `parsedMessage.errors` to subscription handler


bodia-uz:
Duplicate #10

PR #12, opened 2016-09-16
Make the onSubscribe async

rricard:
This is needed if we want to re-extract context from some async store (redis, dataloader, etc...)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
     "url": "git+https://github.com/apollostack/subscriptions-transport-ws.git"
   },
   "dependencies": {
-    "babel-polyfill": "^6.13.0",
     "es6-promise": "^3.2.1",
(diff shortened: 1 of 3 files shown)


rricard:
I can refactor that without the async/await if you don't want the polyfill reinjection!

stubailo:
Yeah can we do it without async/await? Don't want two transpilation steps. I think typescript will have it soon, they said 2.1 last time I checked.

rricard:
@stubailo will do that, I just used it to get it working faster, I prefer the `.then()` notation anyway.

rricard:
@stubailo done! Just a remaining intermittent test fail is making the build red.

helfer:
Thanks @rricard, I think this will be very useful. Can you rebase it so I can merge it?

rricard:
@helfer: done

PR #15, opened 2016-09-19
Use specific lodash functions instead of package

rricard:
Importing the complete lodash package is:

- heavy, we only need two functions here!
- side effects by changing the _ in the global scope (for example, someone using underscore in a legacy backbone app will be affected ;) )

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   "dependencies": {
     "es6-promise": "^3.2.1",
     "graphql-subscriptions": "^0.1.3",
-    "lodash": "^4.15.0",
+    "lodash.isobject": "^3.0.2",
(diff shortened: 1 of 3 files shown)


helfer:
@rricard Thanks! I'll merge it once it's rebased.

rricard:
@helfer done!

helfer:
@rricard ok, can you fix the CI, please. It doesn't pass at the moment.

rricard:
@helfer yep, figured out it would fail while redoing the typing in the other rebase! I went a bit too fast

rricard:
@helfer should pass now!

rricard:
and rerebased again!

rricard:
intermittent ci fail again, see #18

PR #17, opened 2016-09-19
Add an onUnsubscribe hook

rricard:
Enable adding some connection state tracking on the server code.

Here are the signatures now:
- `onSubscribe(parsedMessage, params, subId)`
- `onUnsubscribe(graphqlSubId, subId)`

By tracking the subId we can now identify if a client is connected or not.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -58,6 +58,7 @@ class Server {
   private onUnsubscribe: Function;
   private wsServer: WebSocketServer;
   private subscriptionManager: SubscriptionManager;
+  private emittedUnsubscribe: {[subId: string]: boolean};
 
   constructor(options: ServerOptions, httpServer: HttpServer) {
     const { subscriptionManager, onSubscribe, onUnsubscribe, keepAlive } = options;
@@ -69,6 +70,7 @@ class Server {
     this.subscriptionManager = subscriptionManager;
     this.onSubscribe = onSubscribe;
     this.onUnsubscribe = onUnsubscribe;
+    this.emittedUnsubscribe = {};
 
     // init and connect websocket server to http
     this.wsServer = new WebSocketServer({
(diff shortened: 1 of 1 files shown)


helfer:
@rricard good idea. I can merge it as soon as you add a test for it 👍

rricard:
@helfer , will be a good occasion to test some stuff marked as `TODO` as well

helfer:
I would just write a simple test that checks that onUnsubscribe actually runs when someone calls unsubscribe.

Oh, and one more thing: You should also think about whether onUnsubscribe should run when the connection just closes without the client calling unsubscribe first.

rricard:
@helfer that's what I was worried about indeed, after using this in a custom fork today, I think I'll try some other edge cases as well

helfer:
@rricard do you think this is close to ready to merge? Let me know when you've gotten the tests to pass, and I'll review it again!

rricard:
@helfer I'm still not completely sure the PR is good. I'm trying to trace a rather large leak that may be caused by this. So I want to be careful here, needs more testing.

kesne:
@rricard Have you had any chance to follow-up on this? If not I'd be happy to take a look as well!

rricard:
@kesne Couldn't finally get a reliable way to do that. I went to an activity-based system where I trigger jobs if and only if an user was active in the last hour.

I wonder if the core meteor team that actually worked a lot with websockets could give their feedback on this... I think they solved most of the problems this package is having!

Urigo:
see discussion here - https://github.com/apollostack/graphql-subscriptions/pull/27

rricard:
I'm closing this for now, I don't think this implementation is good. Feel free to get inspiration from it but know it has many shortcomings (we can't really ensure the hook will be called!)

PR #19, opened 2016-09-19
fix(release): fix for typings reference issue

DxCx:


helfer:
Thanks! Can you do a quick check on our other repos to make sure we don't have that same option there? If we do, just file an issue so we can take care of it later.

PR #21, opened 2016-09-21
fixes #20 by checking protocols requested by client before accepting

### LINKED ISSUE #20: server crash with bad client
The server does not correctly handle a client that sends a bad protocol resulting in a server crash. Using GitHunt-API as an example:

```
GitHunt-API/node_modules/websocket/lib/WebSocketRequest.js:289
            throw new Error('Specified protocol was not requested by the client.');
            ^

Error: Specified protocol was not requested by the client.
    at WebSocketRequest.accept (GitHunt-API/node_modules/websocket/lib/WebSocketRequest.js:289:19)
```

Testing using the `wscat` npm module. After launching the server run: `wscat -c ws://localhost:8080`

Issue looks to be that https://github.com/apollostack/subscriptions-transport-ws/blob/master/src/server.ts#L76 blindly accepts all clients.

NeoPhi:


helfer:
Awesome, thanks!

PR #22, opened 2016-09-26
Keep alive socket

rricard:
Some PAAS such as heroku limit the time a socket can stay open without sending anything. By adding the `keepAlive` option in the server's settings, you can now make sure the socket will stay open by sending empty keep alive packets in it from time to time.

`keepAlive` takes an amount of milliseconds as parameter and will yield an empty keep alive message each time this interval goes off.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -7,6 +7,7 @@ import {
   SUBSCRIPTION_START,
   SUBSCRIPTION_SUCCESS,
   SUBSCRIPTION_END,
+  SUBSCRIPTION_KEEPALIVE,
 } from './messageTypes';
 import { GRAPHQL_SUBSCRIPTIONS } from './protocols';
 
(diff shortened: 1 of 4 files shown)


helfer:
@rricard Good idea, thanks!

rricard:
@helfer thanks! I still have some issues with the client, expect a second PR ;)

tailhook:
Maybe I'm duplicating question in #28, but why don't you send regular websocket ping? (It's transparently and automatically responded by browser)

Yogu:
@tailhook Good question, I was wondering the same after I had problems with a client that did not respond to the "ka"-typed frame and the server disconnected the connection after 30 seconds.

Instead of using `keepAlive`, I now use `onConnection` and `onDisconnect` to start and stop an interval that calls `ping` on the websocket implementation. This is is how it's configured in nest.js:

```typescript
const pingIntervals = new WeakMap<object, NodeJS.Timeout>();

@Module({
  imports: [
    GraphQLModule.forRoot({
      typePaths: [__dirname + '/**/*.graphqls'],
      installSubscriptionHandlers: true,
      subscriptions: {
        onConnect(connectionParams, websocket, context) {
          pingIntervals.set(websocket, setInterval(() => websocket.ping(), 30000));
        },
        onDisconnect(websocket, context) {
          const interval = pingIntervals.get(websocket);
          if (interval) {
            clearInterval(interval);
            pingIntervals.delete(websocket);
          }
        }
      },
    })
  ],
})
```

PR #26, opened 2016-09-29
delay the subscribe request to ensure the timeout fires, fixes #18

### LINKED ISSUE #18: Intermittent test/CI fail - Client should throw an error when the susbcription times out
```
1) Client should throw an error when the susbcription times out:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
      at Timeout.<anonymous> (/home/travis/build/apollostack/subscriptions-transport-ws/node_modules/mocha/lib/runnable.js:230:19)
      at tryOnTimeout (timers.js:232:11)
      at Timer.listOnTimeout (timers.js:202:5)
```

CI and local tests fail regularly on this specific test. Maybe we're a bit tight in authorised timeout in the tests.

NeoPhi:
The key change is adding a new test server that delays the response from onSubscribe. It also refactors some duplicated boilerplate code from the other test server instances being created and fixes whitespace in the test method.

### DIFF
--- a/src/test/tests.ts
+++ b/src/test/tests.ts
@@ -22,7 +22,7 @@ import {
   GRAPHQL_SUBSCRIPTIONS,
 } from '../protocols';
 
-import { createServer } from 'http';
+import { createServer, IncomingMessage, ServerResponse } from 'http';
 import SubscriptionServer from '../server';
 import Client from '../client';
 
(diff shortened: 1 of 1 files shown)


helfer:
Thanks, looks good to me!

PR #27, opened 2016-09-29
improve error handling in server and client

NeoPhi:
I noticed that is the `subscriptionManager.subscribe` threw a regular Error object the client received `undefined` despite the message type being `SUBSCRIPTION_FAIL` hence the introduction of `formatErrors`. The server was updated to check for `e.errors` and otherwise create a formatted error object.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -168,9 +168,12 @@ export default class Client {
   }
 
   // ensure we have an array of errors
-  private formatErrors(errors: Array<any>) {
+  private formatErrors(errors: any) {
     if (Array.isArray(errors)) {
       return errors;
+    }
+    if (errors && errors.message) {
+      return [errors];
     }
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -167,4 +167,11 @@ export default class Client {
     return id;
   }
 
+  // ensure we have an array of errors
+  private formatErrors(errors: Array<any>) {
+    if (Array.isArray(errors)) {
+      return errors;
+    }
+    return [{ message: 'Unknown error' }];
```
helfer: Couldn't we at least provide the error's message here?

NeoPhi: Good point. I added support for that and fleshed out the tests a bit more.

helfer:
Thanks, merged!

PR #29, opened 2016-09-30
Don't let transport to fail silently

vlasenko:
Don't let transport to fail silently on general exceptions. Send error message to client instead.


helfer:
@vlasenko Thanks for the PR! I think this was taken care of in #27.

If you think it's been fixed, please close the PR, if not you can rebase it and I'll take another look!

vlasenko:
Yes, I think it should be fixed by #27

PR #30, opened 2016-10-03
Pass WebSocketRequest to onSubscribe

NeoPhi:
To support reading HTTP headers from the web socket request this change passes along the WebSocketRequest to the onSubscribe method as a third parameter.

In the process of documenting the API change I reworked to the README to better reflect the current API.


stubailo:
This looks great - and thanks for updating the readme!

PR #31, opened 2016-10-04
Client reconnection logic

### LINKED ISSUE #4: Reconnect when server closes connection?
If the server closes the connection, we don't try to reconnect and resubscribe, which means that any time the server restarts, all active subscriptions get dropped on the floor.

NeoPhi:
This is an initial approach to handling having the client reconnect and resubscribe after losing its connection to the server. To enable this behavior a new `reconnect: true` option must be passed in to the constructor. Fixes #4 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "url": "git+https://github.com/apollostack/subscriptions-transport-ws.git"
   },
   "dependencies": {
+    "@types/node": "^6.0.38",
     "backo2": "^1.0.2",
     "es6-promise": "^3.2.1",
     "graphql-subscriptions": "^0.1.3",
@@ -31,7 +32,6 @@
     "@types/chai": "^3.4.32",
     "@types/lodash": "^4.14.34",
     "@types/mocha": "^2.2.31",
-    "@types/node": "^6.0.38",
     "@types/sinon": "^1.16.31",
     "@types/websocket": "0.0.30",
     "babel-cli": "^6.11.4",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url": "git+https://github.com/apollostack/subscriptions-transport-ws.git"
   },
   "dependencies": {
-    "@types/node": "^6.0.38",
+    "backo2": "^1.0.2",
     "es6-promise": "^3.2.1",
     "graphql-subscriptions": "^0.1.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "url": "git+https://github.com/apollostack/subscriptions-transport-ws.git"
   },
   "dependencies": {
-    "@types/node": "^6.0.38",
```
helfer: @DxCx moved the node typings here to solve a problem we had with the release, but I can't quite remember what it was. Are you sure we don't need them any more (it seems like a change unrelated to the PR)

src/test/tests.ts
```diff
@@ -352,6 +352,73 @@ describe('Client', function() {
...
+        assert(false);
+      }
+    });
+    const client = new Client(`ws://localhost:${RAW_TEST_PORT}/`, {
+      timeout: 100,
+      reconnect: true,
+      reconnectionAttempts: 1,
+    });
+    setTimeout(() => {
+      expect(client.client.readyState).to.be.equal(client.client.CLOSED);
```
helfer: Is the state otherwise guaranteed to be `CONNECTING`?

package.json
NeoPhi: It seemed strange that when pulling the project downstream that it was installing this as a dependency. I noticed and moved since I was adding the backo2 module. I can revert that change if desired.

src/test/tests.ts
NeoPhi: Other options are `OPEN` or `CLOSING`. It almost felt like we should add lifecycle events to the client that you can listen for which is similar to what socket.io uses.

package.json
DxCx: well, it might be my mistake (that it's dependency and not devDependency.)
i don't think it should be removed but i suggest to move it to devDepenency and make sure package still importable..

NeoPhi:
I've been testing the reconnect logic this afternoon and initial tests are working.

package.json
DxCx: Great! Then we are good as far as i concern

helfer:
@NeoPhi Was that other project a typescript project or a javascript project?

NeoPhi:
I reverted the change to move the @types.

stubailo:
Looks great!

PR #34, opened 2016-10-13
update dependencies and remove unneeded dependencies

NeoPhi:
The primary change was to bump `graphql-subscriptions` up to the latest version which required a minor change in the test suite to account for the new `setupFunctions` syntax.

Since all code is being compiled using `tsc`, babel is not needed to run the tests or coverage.

It looks like there were some dependencies from earlier versions of the project that are no longer referenced so these were removed.

Lastly the types information should only be a `devDependencies` as it isn't needed during runtime.


helfer:
Looks great, thanks @NeoPhi!

PR #42, opened 2016-11-01
fix todo of handling runtime errors in server callback method

NeoPhi:
TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### vNEXT
 
 - ...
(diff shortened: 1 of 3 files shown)


NeoPhi:
@helfer I believe that case is already covered by `should call error handler when graphql result has errors`: https://github.com/apollostack/subscriptions-transport-ws/blob/master/src/test/tests.ts#L239-L264

helfer:
@NeoPhi yes, you're right! :)

Issue #47, opened 2016-11-15
Provide a browser bundle suitable for direct browser consumption


schrepfler:
Noticed the client uses `require` which makes sense if one uses it on node or via browserify/build tools. What if we also expose a browser bundle for the client which already ships everything packaged so that it's possible to reference the library directly via https://unpkg.com?

stubailo:
@schrepfler if you could send a PR with a script that uses Webpack or Browserify to generate a bundle so that we can use `unpkg`, that would be great!

AlexLeung:
I use `unpkg` all the time. This would be really helpful!

schrepfler:
I think this is the basis https://github.com/apollostack/subscriptions-transport-ws/pull/48 but I need some validation as to 1) is the generated bundle consumable from a </script> tag  and will the configuration publish the bundle onto npm or do I need to add more stuff to the package.json (I'm not really a JS dev so all this module mumbo jumbo seems a bit like black magic.)

schrepfler:
Although I think we merged the support for this, I can see https://unpkg.com/subscriptions-transport-ws@0.5.3 contains require statements (which I don't have locally.)
ex. locally the client.js transpiled bundle starts with
```
var SubscriptionsTransportWs =
/******/ (function(modules) { // webpackBootstrap
/******/        // The module cache
/******/        var installedModules = {};

/******/        // The require function
/******/        function __webpack_require__(moduleId) {

/******/                // Check if module is in cache
/******/                if(installedModules[moduleId])
/******/                        return installedModules[moduleId].exports;

/******/                // Create a new module (and put it into the cache)
/******/                var module = installedModules[moduleId] = {
/******/                        i: moduleId,
/******/                        l: false,
/******/                        exports: {}
/******/                };
```

schrepfler:
Actually, there seems to have been a regression, 0.5.2 was generating a different output: https://unpkg.com/subscriptions-transport-ws@0.5.2

schrepfler:
Went through the changes for 0.5.3 and noticed the change in use of unpkg and that I need to explicitly use https://unpkg.com/subscriptions-transport-ws@0.5.3/browser/client.js 
I think this beats the purpose of the "browser" node which should specify the default browser main entry point, if that was kept as-is https://unpkg.com/subscriptions-transport-ws@0.5.3 would be enough to specify?

Urigo:
Fixed in `0.5.1`.

PR #48, opened 2016-11-15
Module browser bundle configuration using webpack

schrepfler:
Only thing I can see is that the bundle is being created but I don't have the skills to really test this (sorry, Java/Scala guy here).
This feature should enable simple CDN-like referencing via https://unpkg.com/ while keeping the publishing to npm as a basis. Related to #47 

TODO:


schrepfler:
To trigger the build one just launches `webpack` after which the browser bundle is created. This command can probably be added in package.json, I guess it depends on your build pipeline. Not sure if I need to add something specific so that npm publish publishes the artifact to npm as well?

schrepfler:
At the moment I'm not sure if the generated bundle is ideal for consumption from </script> tags. Would be nice if a webpack expert suggests what's missing and also confirm will npm publish publish the browser bundle automatically or not?

schrepfler:
I think at the moment the bundle is not properly created, don't merge just yet (and if anyone has suggestions feel free to comment on the PR.)

schrepfler:
I think with this final commit, the generated bundle is usable in a browser via a script tag
```
var client = new subscriptions_transport_ws.Client('ws://localhost:8080/graphql', {timeout: 4000})
```
and 
```
npm pack 
```
generates a tgz which contains it so I suppose this is more/less what we want?

schrepfler:
The only thing which actually comes to mind, if subscriptions_transport_ws imports things from graphql-subscriptions and they finish in the bundle, it might be possible to say that these need to be provided as linked scripts from the outside (declare them external in webpack) but at this point I think it's better for you guys to evaluate.

schrepfler:
Any feedback on this would be nice?

helfer:
@schrepfler sorry, this fell off of my radar for a while. I think we should definitely implement it, but it looks like you're compiling things twice now (once with TS once with webpack). Couldn't webpack just use the output from TS and bundle for the browser after that?

@rricard I think you know this library (and also js in general 😄 ) pretty well. Could you take a quick look at this PR and help @schrepfler where necessary? That would be awesome!

rricard:
@helfer thanks for directing me to this issue.

@schrepfler I'm going to do a proper review of your PR.

package.json
```diff
@@ -24,7 +26,8 @@
     "testonly": "mocha --reporter spec --full-trace ./dist/test/tests.js",
     "coverage": "node ./node_modules/istanbul/lib/cli.js cover _mocha -- --full-trace ./dist/test/tests.js",
     "postcoverage": "remap-istanbul --input coverage/coverage.raw.json --type lcovonly --output coverage/lcov.info",
-    "prepublish": "npm run compile"
+    "start": "webpack --config webpack.config.js",
```
rricard: Don't call this script start, something more meaningful: `browser-compile`

rricard: `webpack.config.js` will be called by default so just put `"webpack"` in the script invocation

package.json
```diff
@@ -24,7 +26,8 @@
...
+    "prepublish": "npm run compile && npm run start"
```
rricard: Instead of `start`, put `browser-compile`

webpack.config.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  context: __dirname + "/src",
+  entry: './index.ts',
+  output: {
+    path: __dirname + '/browser',
+    filename: 'bundle.js',
+    library: 'subscriptions_transport_ws'
```
rricard: Usually exported libs are in pascal case: `'SubscriptionsTransportWs'`, please do that!

webpack.config.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  context: __dirname + "/src",
```
rricard: Use [`path.join`](https://nodejs.org/api/path.html) instead of a concatenation

webpack.config.js
```diff
@@ -0,0 +1,18 @@
...
+  entry: './index.ts',
+  output: {
+    path: __dirname + '/browser',
```
rricard: Use [`path.join`](https://nodejs.org/api/path.html) instead of a concatenation

webpack.config.js
```diff
@@ -0,0 +1,18 @@
...
+    filename: 'bundle.js',
+    library: 'subscriptions_transport_ws'
+  },
+  resolve: {
+    extensions: ['', '.webpack.js', '.web.js', '.ts', '.js']
```
rricard: Why not just `['.ts', '.js', '.json']` ?

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ### vNEXT
 
-- ...
+- Webpack configuration to export the generated code into a browser compatible bundle.
```
rricard: Nitpick: reference this PR number in the changelog

rricard:
@helfer I'm concerned about the dependency on `graphql-subscriptions` wouldn't it be better to put it as a peer dependency? Normally, the browser bundle shouldn't touch it but even with that, having it as a dependency in node is not ideal... And having it as a peer would let me shim it in the webpack config.

rricard:
@schrepfler On our side we should only expose the client module, so your entry point should only be: `src/client.ts`, that way we avoid the hassle with `graphql-subscriptions`

helfer:
@rricard thanks a lot for tuning in here! :heart: We can definitely (and probably should) make graphql-subscriptions a peer dependency.

package.json
```diff
@@ -10,8 +10,10 @@
   "dependencies": {
     "backo2": "^1.0.2",
     "graphql-subscriptions": "^0.2.0",
```
rricard: Ok, so while we're here, can you move this one in `devDependencies` and `peerDependencies`.

schrepfler:
Went over the suggestions, only problem was with `['.ts', '.js', '.json']` I had to add '' to the list.

schrepfler:
Ah, I also bumped a bit the libraries, but notably typed-graphql and graphql to get in line with the rest of the apollo stack. Anyhow, merry xMas!

schrepfler:
Would the addition of `graphql-subscriptions` to peerDependencies mean that we would also need to create a browser bundle for it as well (and the other deps?)

Akryum:
@rricard If a module is never imported in the code, webpack will not include it in the bundle.

package.json
```diff
@@ -24,29 +25,36 @@
...
     "tslint": "^3.13.0",
-    "typed-graphql": "^1.0.1",
-    "typescript": "^2.0.0",
-    "typings": "^1.3.2"
+    "typed-graphql": "^1.0.2",
+    "typescript": "^2.0.3",
+    "typings": "^1.3.2",
+    "webpack": "^1.14.0",
+    "graphql-subscriptions": "^0.2.2"
+  },
```
rricard: Updating all of those packages broke the CI.

schrepfler: Only these three all or should I revert all of them?

rricard: all

rricard: remap-istanbul particularly since it broke the build

schrepfler:
Reverted library updates from devDependencies (except the discussed peer deps stuff.) Any thoughts around json-loader and ts-loader in the dependencies, should they be runtime dependencies or should they be in the devDependencies section?

schrepfler:
@rricard can you please review the reversal of the change and comment on the runtime vs devDependencies wrt json and ts-loaders?

schrepfler:
Shall we try to ship it and then try to catch up with the rest of the apollo js stack? 😀🚀🛫🍻

Urigo:
Maybe we can use the changes from this PR: https://github.com/apollographql/subscriptions-transport-ws/pull/53 ? the package now compiles and create two entry points (one for `client.ts` and one for `server.ts`), and the `client` file does not have an external dependencies. 
@rricard , maybe we can use this to avoid two compilations?

schrepfler:
Hi @Urigo, not sure if I understand, are you saying #53 can deprecate this PR? 
The rationale of #48 is more to create a browser bundle which can be directly used via unpkg.com once the library gets published onto npm.

Urigo:
@schrepfler I meant that instead of creating a new Webpack config file and compile the package for browser, we might be able to do the same bundle process from the compiled files, since they're no longer coupled.
Also, can you please rebase this PR? thanks.

schrepfler:
Not sure what's going on, the rebase went into a loop, so tried to do the merge manually.

schrepfler:
Somehow the new versions popped back into package.json. Hopefully it's good now.

Urigo:
@schrepfler , thanks for the rebase. What about using the compiled client file, and then just bundling it?

schrepfler:
I think this is the case, the only thing webpack is bundling is the client.ts? https://github.com/apollographql/subscriptions-transport-ws/pull/48/files#diff-11e9f7f953edc64ba14b0cc350ae7b9d

schrepfler:
Also, please consider this has been submitted on the 15th November 2016 and JS is not my main skill, can we wrap it up? I don't mind if you use mine bundle or if you have your bundle. 
All it has to provide is:
* the package.json browser element pointing to the right bundle and 
* that bundle needs to be consumable directly in a browser context (so I guess no `requires`.) This bundle needs to target the browser, not node.
* bonus: also export the helper fn? (I didn't do anything around that, personally I'd move it into the client bundle)

Urigo:
@schrepfler sorry it took a long time, the other PR was pretty big and I was working on it while implementing subscriptions on one of our internal apps.
So about your last comment, the first and last points should be done with the new version.
So now what's left is just to expose the compiled client code without require.
You can update this PR or create another one

schrepfler:
I think this is the case, I wrap client.ts which is then exposed as SubscriptionsTransportWs in the global scope if added to the page. *I think*

schrepfler:
I just realised you're the same guy from the JSJabber podcast episode I just listened, nice one! :D

schrepfler:
Any feedback in order to understand further course of actions?

schrepfler:
I'll close this PR and start over.

Urigo:
thanks again @schrepfler , linking the merged one if someone will stumble upon this one - https://github.com/apollographql/subscriptions-transport-ws/pull/77

PR #71, opened 2017-02-09
only attempt reconnect on a closed connection, fix for #70

### LINKED ISSUE #70: Is the reconnect backoff working?
I just upgraded to 0.5.0 to get reconnect working. It seems to really hammer the browser when I kill my graphql server. It gets so bad after a couple minutes that I can't even kill the tab in Chrome.

Is the backoff delay working currently?

NeoPhi:
Fixes #70 

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### vNEXT
 
 - ...
+- Only attempt reconnect on closed connection. Fixes [Issue #70](https://github.com/apollographql/subscriptions-transport-ws/issues/70)
 
 ### 0.5.0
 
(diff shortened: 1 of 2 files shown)


helfer:
@NeoPhi I'll merge this and release a new version now since it's an issue that can affect production apps. I'd still like to know more about the uncaught errors in tests that you mentioned in #70 though?

NeoPhi:
The test `'should stop trying to reconnect to the server'` shutdowns the server while the client is still connected. The Node implementation of `WebSocket` from the `ws` package triggers an error at this point. If the `onerror` is not defined in the client it results in a `Uncaught Error: socket hang up` being reported. Since I'm not confident that the native browser implementation of `WebSocket` doesn't suffer the same issue, I suspect it is best to leave the no-op handler in place.

PR #73, opened 2017-02-10
Create index.ts

BernhardRode:
@angular/cli is not able to resolve the package.json 'browser' attribute. Maybe it would be better to have this simple index.ts file and change the package.json to point to it.

at the moment i have to use 

import { SubscriptionClient, addGraphQLSubscriptions } from 'subscriptions-transport-ws/dist/client';

to get import working in latest cli.

TODO:

- [ ] Rebase your changes on master so that they can be merged easily

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,0 +1,5 @@
+export * from './client';
+export * from './server';
+export * from './helpers';
+export * from './messageTypes';
+export * from './protocols';


helfer:
Thanks @BernhardRode 

@kamilkisiela @Urigo any advice on this?

kamilkisiela:
```ts
import { something } from 'some-package';
```

I'm not quite sure but `some-package` points to a file specified in the `main` field.
So `some-package` points to `some-package/dist/server.d.ts`.

Are you sure it works when you just create `index.ts`?

Urigo:
Thanks @BernhardRode . merged. we will also fix the issue with the `package.json` soon.

Issue #75, opened 2017-02-15
authorizing subscriptions -- set a default context for client


srtucker22:
My goal is simply to validate whether a user is authorized to subscribe to a given channel. For example, if I was creating a group chat app, only members of a group would be able to subscribe to `onMessageAdded(groupId: Int!)` for that groupId.

From my understanding from digging into the code...

On the server, I can accomplish allowing/denying subscriptions via onSubscribe with a given context:

```
onSubscribe(parsedMessage, baseParams, connection) {
   // do stuff with baseParams.context to validate subscription
},
```

On the client, I'm trying to figure out a way to set up a default context for all subscriptions, similar to how `networkInterface` exposes this with middleware. For example, using `networkInterface`, we have a redux store that sets a jwt token on authorization headers. We send out queries with auth headers if the jwt is set, and on the server side, we validate the jwt token and pass the validated user into `context` for the resolvers to consume.

```
// middleware for requests
networkInterface.use([{
  applyMiddleware(req, next) {
    if (!req.options.headers) {
      req.options.headers = {};
    }

    // get the authentication token from local storage if it exists
    const jwt = store.getState().auth.jwt;
    if (jwt) {
      req.options.headers.authorization = `Bearer ${jwt}`;
    }
    next();
  },
}]);
```

I'd imagine something similar would/should be possible with SubscriptionManager, where we either (1) apply context via a setupFunction using `next` or a Promise, or (2) expose some public function that updates the default context for all subscriptions. This seems particularly important if we're using `subscribeToMore` as I don't see an obvious way to pass context with this call.

Any help or suggestions appreciated!

hmaurer:
I think this is solved by `connectionParams`? c.f. https://github.com/apollographql/subscriptions-transport-ws/pull/53

srtucker22:
Nope, I don't think it solves the challenge but please correct me if I'm wrong.

As I understand it:
`connectionParams` on the client deals with `onConnect` so when you make a socket connection. But won't affect anything during `client.subscribe()`. A common use case where this matters would be letting anyone make a connection, but only let users subscribe to events with an auth context. It would be nice to pass all subscribe requests through a function that attaches auth context like a JWT token.

[...]

PR #78, opened 2017-02-20
onSubscribe and use middleware to modify runtime SubscriptionOptions

srtucker22:
TODO:


srtucker22:
This is related to #75 -- while it doesn't create a default context, it instead allows you to override the `SubscriptionOptions` passed into `subscribe` much like `applyMiddleware` overrides the request in `networkInterface`. 

It's a very basic solution to the broad challenge I'm facing -- having some macro controls over subscription creation with `apollo-client`. A typical example of this is setting authorization context for new subscriptions after a connection is made. We want to make a connection right away without worrying about authorization, and then run all queries through the same logic to establish auth context when subscribing to events.

NeoPhi:
I feel this would be better handled in the glue layer between the `networkInterface` and the `wsClient`. An alternative implementation of [`addGraphQLSubscriptions`](https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/helpers.ts) with hooks would allow you to modify the options being passed to `subscribe`.

srtucker22:
so something like?

```js
import { SubscriptionClient } from './client';
import { print } from 'graphql-tag/printer';

// Quick way to add the subscribe and unsubscribe functions to the network interface
export function addGraphQLSubscriptions(networkInterface: any, wsClient: SubscriptionClient): any {
  return Object.assign(networkInterface, {
    subscribe(request: any, handler: any): number {
      if(networkInterface.onSubscribe) {
        request = networkInterface.onSubscribe(request);
      }
      
      return wsClient.subscribe({
        query: print(request.query),
        variables: request.variables,
        context: request.context,
      }, handler);
    },
    unsubscribe(id: number): void {
      wsClient.unsubscribe(id);
    },
  });
}
```

At the very least I think the subscribe function should be able to pass context.

Just to be clear about the use case I'm running into, which I don't think is that uncommon:
I want to be able to login/logout and have that auth context passed to my subscriptions. 
If using `apollo-client` or `react-apollo`, there is no way of affecting context.
you can pass context when first connecting to socket, but it would be inefficient to open/close connection every login/logout.

Even if we exposed context in `subscribe`, you wouldn't want to have to explicitly pass context in every `subscribe` or `subscribeToMore` call if context is something high level like a JWT token. I also imagine there are other things someone might need to do before every `subscribe` call just like it's import to have middleware before a gql request sent with `networkInterface`.

Urigo:
@srtucker22 looks really good and I agree that this is a common use case that would be beneficial to support.

If we choose to support it then it will be easier then @NeoPhi solution overriding the subscribe method.

Maybe a better way of supporting it would be to add another client life-cycle hook called `onSubscribe` and this transport will call this hook right after [this](https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L115) and extend the `options` with the return value of the hook.

That means that the hooks can return only the part of the `options` that it want to override or extend and not making it necessary to build the whole `options` object again from scratch. 

In the client there is already lifecycle hook event emitter we can use for that, here is an example - https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L147-L157

Would love to hear both of your thoughts..

Urigo:
also, maybe the best thing will be to make it as similar as possible to that existing API - http://dev.apollodata.com/core/network.html#networkInterfaceMiddleware

srtucker22:
Hey @Urigo, good to hear from you!

I'm all for `networkInterfaceMiddleware` style middleware. Are you thinking it would look something like this?
```js
SubscriptionClient.use([{
  applyMiddleware(options, next) {
    // modify options for SubscriptionClient.subscribe here
    next();
  }
}]);
```
My only concern with adding middlware this way (as it's written in `apollo-client`) is it introduces promises and async. I like how in the current setup, `subscribe` is a synchronous function that returns the subscription ID or throws. I think it would be best to still return an ID synchronously, but I wouldn't want to use promises in the background, especially when things throw errors. We could restrict middleware funcs to sync only, but that kind of defeats the purpose of using `next()`. Maybe can be overcome with iterators and generators?

srtucker22:
ok so this latest push will use either `client.onSubscribe` or `client.use`, but both are syncronous and so very basic.

i can forsee use cases where someone would like to run an async function to modify options before subscribing, and this doesn't address that. but anything with async middleware before subscribe will essentially demand breaking changes whereas these changes don't.

happy to modify more if we want to go down the breaking changes path :)

Urigo:
Thank you for your recent changes.
I think that adding support for async is important, and I think it might be simpler. 
Can you please explain why the changes will break the API?

Thanks @srtucker22

srtucker22:
so i guess if you do it this way (latest commit) the changes aren't breaking per se -- before this change you already had to catch errors (1) when subscribing and (2) in the handler, but now some of those errors (like having a query) will also be caught in the handler where they weren't before, like if your middleware accidentally deleted the query.

however, this code has an issue --> if you call `client.unsubscribe` before `client.subscribe` finishes (as was written in a few tests), you will not unsubscribe because unsub will run first in the queue / run before the `graphqlSubId` is created and is necessary to call `unsubscribe` on the server. not sure how you want to tackle this --> maybe something like shift unsubs and corresponding subs around in the queue so they're in the right order?

apologies for being touch and go. i've been away in Bali these past few weeks and will be traveling for another couple months or so. just want to give you a heads up i might not respond for a few days  in between coms

srtucker22:
ok, this version tracks unsubs and waits to sendMessage for unsubscribe if it comes before sendMessage for init for a given sub. doesn't feel elegant, but gets the job done

srtucker22:
a little cleaner now. lemme know what you think

in tokyo now so better internet sitch

src/client.ts
```diff
@@ -110,37 +116,38 @@ export class SubscriptionClient {
     this.client.close();
   }
 
-  public subscribe(options: SubscriptionOptions, handler: (error: Error[], result?: any) => void) {
-    const { query, variables, operationName, context } = options;
+  public subscribe(opts: SubscriptionOptions, handler: (error: Error[], result?: any) => void) {
+    // this.eventEmitter.emit('subscribe', options);
+    this.checkSubscriptionParams(opts, handler);
```
NeoPhi: I would just rely on the `this. applyMiddlewares` handling instead of checking it twice. This also avoids the case that sometimes the handler is called with the error and not.

src/client.ts
```diff
@@ -110,37 +116,38 @@ export class SubscriptionClient {
...
-    }
+    const subId = this.generateSubscriptionId();
 
-    if (!handler) {
-      throw new Error('Must provide `handler` to subscribe.');
-    }
+    this.applyMiddlewares(opts).then(options => {
+      try {
+        this.checkSubscriptionParams(options, handler);
+      } catch (e) {
```
NeoPhi: Feels like this should just be a `catch` on the promise to catch any issues.

src/client.ts
```diff
@@ -110,37 +116,38 @@ export class SubscriptionClient {
...
+        handler([e]);
+        this.unsubscribe(subId);
```
NeoPhi: We should unsubscribe before calling the handler.

src/client.ts
```diff
@@ -110,37 +116,38 @@ export class SubscriptionClient {
...
+      if (this.waitingUnsubscribes[subId]) {
+        delete this.waitingUnsubscribes[subId];
         this.unsubscribe(subId);
       }
-    }, this.subscriptionTimeout);
+
+      setTimeout( () => {
+        if (this.waitingSubscriptions[subId]) {
+          handler([new Error('Subscription timed out - no response from server')]);
+          this.unsubscribe(subId);
```
NeoPhi: We should unsubscribe before calling the handler.

src/client.ts
```diff
@@ -165,9 +172,14 @@ export class SubscriptionClient {
   }
 
   public unsubscribe(id: number) {
+    if (!this.subscriptions[id] && !this.waitingSubscriptions[id]) {
+      this.waitingUnsubscribes[id] = true;
```
NeoPhi: I'm not sure what `waitingUnsubscribes` is giving us. Each `subscribe` call is going to generate a new id and I don't see this being used elsewhere.

src/client.ts
srtucker22: so if we change this, it will break the following tests as they currently stand:

1. should throw an exception when query is not provided
2. should throw an exception when query is not valid

because those errors currently throw outside the handler

i'm all for deferring all error handling to the handler (aside from throwing if there isn't a handler), but was trying to make this feature without making breaking changes. but i agree it's cleaner -- so if you confirm this is the route to go, i'm happy to make the change

helfer:
@Urigo could you take another look at this? Thanks! 🙂

Urigo:
hi, I'm sorry for the delay but we have a huge important change [here](https://github.com/apollographql/subscriptions-transport-ws/pull/108) that should land soon and after that I will address the rest of the PRs.
@srtucker22 also be aware of that change because probably after there will be a need for rebase and changes

srtucker22:
thanks for the heads up -- looks cool. also looks like i'm gonna have to rewrite my subscription tutorial :)

src/client.ts
```diff
@@ -177,6 +189,59 @@ export class SubscriptionClient {
     });
   }
 
+  public applyMiddlewares(options: SubscriptionOptions): Promise<SubscriptionOptions> {
+    return new Promise((resolve, reject) => {
+      const queue = (funcs: MiddlewareInterface[], scope: any) => {
+        const next = () => {
```
giautm: Can be reject with `next(error)`?

giautm: Following lines should be change to:
```
      const queue = (funcs: MiddlewareInterface[], scope: any) => {
        const next = (error) => {
          if (error) {
            reject(error);
          } else {
            if (funcs.length > 0) {
              const f = funcs.shift();
              if (f) {
                f.applyMiddleware.apply(scope, [options, next]);
              }
            } else {
              resolve(options);
            }
          }
        };
        next();
      };
```

And [this](https://github.com/srtucker22/subscriptions-transport-ws/blob/b5756399525aed21ef355aa04c5870ccd3d4b6ff/src/client.ts#L149) line update to:
```
}).catch(e => handler([e]));
```

giautm:
Is this ready for merge, wait too long. :(

srtucker22:
SO, here is my proposal with the latest:

Everything works, but I've changed up a few tests since we're in the spirit of major code overhaul:

1. All `query`, `variables`, and `operationName` related errors are thrown in the handler. Since we're giving users the option to use asynchronous middleware to change all of these options, they should really only get checked once they have been updated. Plus, I think it's much more intuitive that all errors get handled by the handler! (changed tests for throwing exceptions to queries to get handled by handler)

2. Because `client.subscribe` returns the `opId` synchronously, but sends the first message asynchronously (after middleware), there is a period in between where the user might try and unsubscribe before the first message is sent. I think the most reasonable thing to do if someone tries to unsubscribe before they've ever actually sent a subscription message to the server is to just cancel the subscription message altogether. In this scenario, the `GQL_START` message won't make it to the queue (1st changed test), and on the server, `onUnsubscribe` won't get triggered (2nd changed test).

Let me know what's good!

Urigo:
@srtucker22 thank you so much for an amazing work and your patience keeping this PR updated.

srtucker22:
no worries!

src/server.ts
```diff
@@ -423,7 +424,7 @@ export class SubscriptionServer {
               query: parsedMessage.payload.query,
               variables: parsedMessage.payload.variables,
               operationName: parsedMessage.payload.operationName,
-              context: Object.assign({}, isObject(initResult) ? initResult : {}),
+              context: Object.assign({}, isObject(initResult) ? initResult : {}, parsedMessage.payload.context),
```
DxCx: Why is context sent over the network?
@urigo @helfer @stubilo - Thats a serious security issue, you shall not release the package with that

quadsurf:
This is still not working for me... I'm still getting "Insufficient Permissions" when I have permissions enabled... however, it works fine when permissions are disabled. I have even hard coded in the server auth token, and it still doesn't work.

There seems to be some property naming inconsistency in this library:
- some use:
connectionParams: {
    Authorization: getToken()
  }
- some use:
connectionParams: {
    authToken: getToken()
  } 

**So is the correct property name "Authorization" or "authToken"?** Either way, I've tested both, and both fail.

HERE ARE MY VERSIONS
react-native@latest
subscriptions-transport-ws@0.8.2
react-apollo@1.4.8

HERE IS MY ERROR
Unhandled GraphQL subscription error Array [
  Object {
    "code": 3008,
    "locations": Array [
      Object {
        "column": 7,
        "line": 4,
      },
    ],
    "message": "Insufficient Permissions",
    "path": Array [
      "User",
      "node",
      "id",
    ],
    "requestId": "subscription:cj6...:1",
  },
  Object {
    "code": 3008,
    "locations": Array [
      Object {
        "column": 7,
        "line": 5,
      },
    ],
    "message": "Insufficient Permissions",
    "path": Array [
      "User",
      "node",
      "type",
    ],
    "requestId": "subscription:cj6...:1",
  },
]

ISSUES I'VE REVIEWED
https://github.com/apollographql/subscriptions-transport-ws/issues/171#issuecomment-316376468 (the one recommended by @lewisblackwood)
https://github.com/graphcool/feature-requests/issues/317
https://github.com/apollographql/subscriptions-transport-ws/issues/102

PR #83, opened 2017-02-27
fix url websocket should prefix 'ws'

kenshero:
// Create WebSocket client prefix url should be 'ws'


kenshero:
OK. I sign the Meteor Contributor already.

NeoPhi:
Thank you very much for the update to the documentation!

PR #84, opened 2017-03-01
Allow data and errors in payload of SUBSCRIPTION_DATA

timsuchanek:
Hi, at Graphcool, we have the use case, that sometimes we want to send `data` AND `errors` on the `payload` of `SUBSCRIPTION_DATA` at the same time.

This is especially needed for our permissions system, where it could be that the requesting client is only allowed to see some fields on a query. In that case we don't want to disable all other fields, so we send the allowed fields and an error that explains why some fields are null.

This PR just makes sure, that if there is the `errors` property, `data` is not just nulled, but is actually checked if there is data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### vNEXT
 
+### 0.5.4
+- Allow data and errors in payload of SUBSCRIPTION_DATA [PR #84](https://github.com/apollographql/subscriptions-transport-ws/pull/84)
+
 ### 0.5.3
 - Fixed a bug with `browser` declaration on package.json ([Issue #79](https://github.com/apollographql/subscriptions-transport-ws/issues/79))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 # Changelog
 
 ### vNEXT
-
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ### vNEXT
 
+### 0.5.4
```
NeoPhi: These should just be placed under vNEXT as multiple different merge requests may get bundled into the next release.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "subscriptions-transport-ws",
-  "version": "0.5.3",
+  "version": "0.5.4",
```
NeoPhi: No need to change this in the MR.

src/client.ts
```diff
@@ -306,7 +306,8 @@ export class SubscriptionClient {
           if (parsedMessage.payload.data && !parsedMessage.payload.errors) {
```
NeoPhi: Probably cleaner to conditionally set `playloadData` and `playloadErrors` and have only a single call to `this.subscriptions[subId].handler`.

timsuchanek:
Thanks for the feedback, good points, just updated the MR

PR #85, opened 2017-03-01
Send INIT before SUBSCRIPTION_START

timsuchanek:
screen shot 2017-03-01 at 4 56 37 pm
Currently, the `SUBSCRIPTION_START` is being sent before the `INIT` message.
In our case this results in `subscriptions-transport-ws` rejecting the `init_success` of the server.
This only happens in rare cases and I could barely reproduce this.
The environment that caused this problem looked like this:
 - Slow connection
 - Having 2 subscriptions in one application
 - Using the `react-apollo` client
I'm not sure if the apollo client is responsible for the weird early `subscription_end`, but this is already a starting point.

Unfortunately, the test I added breaks 3 other Server related tests. Seems some async/concurrency issue, but I couldn't find out what exactly is causing it.
If it is ok for you, I also just can remove this test, as all other tests then still are green.
This is based on PR #84, so you might first merge PR #84.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -306,7 +306,8 @@ export class SubscriptionClient {
           if (parsedMessage.payload.data && !parsedMessage.payload.errors) {
               this.subscriptions[subId].handler(null, parsedMessage.payload.data);
           } else {
-            this.subscriptions[subId].handler(this.formatErrors(parsedMessage.payload.errors), null);
+            const payloadData = parseMessage.payload.data || null
+            this.subscriptions[subId].handler(this.formatErrors(parsedMessage.payload.errors), payloadData);
           }
           break;
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -306,7 +306,7 @@ export class SubscriptionClient {
           if (parsedMessage.payload.data && !parsedMessage.payload.errors) {
               this.subscriptions[subId].handler(null, parsedMessage.payload.data);
           } else {
-            const payloadData = parseMessage.payload.data || null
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -243,6 +243,9 @@ export class SubscriptionClient {
       this.eventEmitter.emit(isReconnect ? 'reconnect' : 'connect');
       this.reconnecting = false;
       this.backoff.reset();
+      // Send INIT message, no need to wait for connection to success (reduce roundtrips)
```
NeoPhi: If I isolate this MR to just moving where the INIT message is fired I'm not seeing any test failures.

src/test/tests.ts
```diff
@@ -213,6 +216,40 @@ describe('Client', function () {
...
+      connection.on('message', (message: any) => {
+        const parsedMessage = JSON.parse(message);
+        // mock server
+        if (parsedMessage.type === INIT) {
+          connection.send(JSON.stringify({type: INIT_SUCCESS, payload: {}}));
+          initReceived = true
+        }
+        if (parsedMessage.type === SUBSCRIPTION_START) {
+          expect(initReceived).to.be.true
+          done()
```
NeoPhi: Need to setup the client as a function variable and include a `client.unsubscribeAll();` prior to calling `done()` otherwise the client will stick around and try to reconnect.

timsuchanek:
Thanks for the tip, just calling `client.unsubscribeAll()` right before calling done did the trick.

PR #86, opened 2017-03-02
Update README.md

timsuchanek:
Add `payload` as it is already sent by the client to the `INIT` message.


NeoPhi:
Thank you for the update!

PR #96, opened 2017-03-11
Add a new GraphiQL fetcher

Urigo:
Add a new GraphiQL fetcher that overrides the default one so we can easily add GraphQL subscriptions support to GraphiQL


src/graphiql-fetcher.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const graphQLFetcher = (subscriptionsClient: SubscriptionClient, fallbackFetcher: Function) => {
+  let activeSubscriptionId: number | null = null;
+
+  return (graphQLParams: any) => {
+    if (subscriptionsClient && activeSubscriptionId !== null) {
+      subscriptionsClient.unsubscribe(activeSubscriptionId);
+    }
+
+    if (subscriptionsClient && graphQLParams.query.startsWith('subscription')) {
```
NeoPhi: This seem like a brittle way to test is it is a subscription.

src/graphiql-fetcher.ts
```diff
@@ -0,0 +1,32 @@
+import { SubscriptionClient } from './client';
+
+export const graphQLFetcher = (subscriptionsClient: SubscriptionClient, fallbackFetcher: Function) => {
+  let activeSubscriptionId: number | null = null;
+
+  return (graphQLParams: any) => {
+    if (subscriptionsClient && activeSubscriptionId !== null) {
```
NeoPhi: While it would require GraphiQL changes to support stopping a subscription, having multiple active subscriptions would be ideal.

src/graphiql-fetcher.ts
Urigo: @NeoPhi Agreed.
I thought to try to do it without adding a dependency to grpahql.js...
But can definitely change that.  what do you think would be better?

src/graphiql-fetcher.ts
Urigo: mmmm because this is a temporary change until [this](https://github.com/graphql/graphiql/pull/340) PR would be merged, it's hard for me to think how keeping multiple subscriptions would look nice in the current GraphiQL UI.
Think that you would have a new subscription but would still keep getting new results of the previous subscriptions.  probably would be confusing.
But maybe it would be an interesting thing to think about a better UI for a deeper change in GraphiQL itself.
Do you still think we should leave the previous subscriptions active or leave it like it is now?

src/graphiql-fetcher.ts
helfer: I agree with @NeoPhi. Instead think you can do pretty much what this does:

https://github.com/apollographql/graphql-anywhere/pull/44/files#diff-52bc6e1910f2b72f10d27006c255168fR55

If you don't want to add the graphql-js dependency, you can write that in the comment there and just use `any` for all the types. But you should definitely respect the operationName as well. GraphiQL is one of the few cases where it's actually useful for people to have many definitions in a document but only execute one. This GraphQL fetcher should support that.

Urigo: GraphiQL exposes `string` to the fetcher, and not GraphQL parsed object (`DocumentNode`). So we have two options:
1. Add dependency for `graphql`, parse the string from GraphiQL and use it to check for subscriptions operations. 
2. Leave as is - maybe look for a substring instead only `startsWith`, or maybe other logic you think is better.

@NeoPhi @helfer what do you think is the preferred solution? or maybe I'm missing a third one?

helfer: GraphiQL already comes with graphql. Is there a way to get it from that bundle? If not, I would suggest this:

1. if `operationName` is set, look for a match for `subscription <operationName>`.
2. if `operationName` is not set, check if subscription is present somewhere before the first `{` in the string.

I think we should only send it over the subscription transport if we're very confident that a subscription should happen, because the error messages from subscriptions aren't as easy to understand as the ones that are returned from the normal graphql executor.

NeoPhi: As this is for GraphiQL I don't think the added weight of pulling in GraphQL is too much of a concern.

Urigo:
@helfer @NeoPhi 
I moved the fetcher implementation into a new package: https://github.com/apollographql/GraphiQL-Subscriptions-Fetcher 
This way, I added `graphql` dependency (and use it for parsing) only for the fetcher package (which loaded only for GraphiQL).

helfer:
Looks good to me!

PR #98, opened 2017-03-16
Use print from graphql directly instead of from graphql-tag

xavcz:
TODO:
---

What's up?
---
Hey! While updating `meteor/apollo` to use this package (yay! :rocket:), we encountered the warning about deprecation of `print` from `graphql-tag` (https://github.com/apollographql/meteor-integration/issues/87).

Following discussions from https://github.com/apollographql/graphql-tag/issues/54 & others PRs/issues related, I've made the change & I've tried it locally: tests pass and it works.

I assumed that `graphql` should be then in `dependencies` more than `devDependencies`, right?


Urigo:
Cool, thanks @xavcz

kachkaev:
Would it be possible to release a new version of `subscriptions-transport-ws` guys? Both `0.5.4`  and `0.5.5-alpha.0` seem to be causing the warning in the console:

```
Warning - the `printer` exports from `graphql-tag` will be removed in the next major version.
See https://github.com/apollographql/graphql-tag/issues/54 for more information.
```

Anthonyzou:
Yeah if you can release a new version that would be great.

Urigo:
@Anthonyzou just released `0.5.5`.   sorry for the delay

Issue #99, opened 2017-03-20
Missing EventEmitter on recent RCs of @angular/cli 


kamilkisiela:
An error:
> eventemitter3_1.EventEmitter is not a constructor

It happens inside a project that has been created by `@angular/cli@^1.0.0-rc.1`.

---

In [client.ts](https://github.com/apollographql/subscriptions-transport-ws/blob/b0c71084e88c11ddfb8cefb1d7996e15507bf808/src/client.ts#L2), we've got this:

```ts
import { EventEmitter } from 'eventemitter3';
```

In [eventemitter3/index.js](https://github.com/primus/eventemitter3/blob/e5ba1e240edfb4f3a4177c592ce77cb949b9aadb/index.js#L304):

```ts
EventEmitter.EventEmitter = EventEmitter;
export default EventEmitter;
```

Webpack creates a bundle file that contains that:

```js
var _eventemitter3 = require('eventemitter3');

// inside SubscriptionClient of `subscriptions-transport-ws`

this.eventEmitter = new _eventemitter3.EventEmitter();
```

The thing is, `_eventemitter3` has no `EventEmitter` property assigned. It's missing...

```js
console.log(_eventemitter3); // EventEmitter class
console.log(_eventemitter3.EventEmitter); // undefined
```

When you look up for the `eventemitter3` package inside the bundle file, there's no `EventEmitter.EventEmitter` but there's another property available that has been assigned the same way (`EventEmitter.prefixed`). It somehow removes the assignment from the code.

Weird...

kamilkisiela:
Continuing apollographql/GitHunt-Angular#100

jmarlow4:
Any progress on this?

jhartma:
having the exact same issue

tobiaghiraldini:
Same issue also here. even with subscriptions-trasport-ws@0.5.5. I also tried to upgrade the eventemitter3 dependency with no luck.
this issue is preventing to setup a ws communication with opengraph servers. Is there any workaround at least?

nakedcity:
check: npm ls eventemiter3
packages are using the eventemiter3 e.g karma uses 1.x
solution: npm i eventemiter3 --save

jhartma:
thanks, that works

amsterdam80:
you welcome

Urigo:
Fixed in `0.6.0`.

Issue #95, opened 2017-03-10
onDisconnect: How to track user?


myyellowshoe:
I'd like to figure out how to run some commands for a specific user when they disconnect. Is there something like a state that is stored to be able to reference within the onDisconnect hook?

NeoPhi:
The same `webSocket` instance is passed to all of the `on*` callbacks. You could use that to track custom state in a `Map()` or possibly on the webSocket itself.

myyellowshoe:
Yeah I noticed that, I was wondering if I could stick some data in there, was thinking it would probably be just easiest to do it that way. Is there a reason I'd use one or the other? 
Also how might I add data to the websocket? Any articles/ reference code you might be able to point me to?

Thanks so much!

NeoPhi:
Using an external [Map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) would ensure that you don't accidentally overwrite any state on the `webSocket` instance. Basic pattern would be:
```
onConnect(connectionParams, webSocket) {
  clients.set(webSocket, {custom: 'data'});
}
onDisconnect(webSocket) {
  const data = clients.get(webSocket);
  // use data as needed
}
```

arendjr:
To add to this, I think it would be valuable if the context properties which are returned from `onConnect()` which be passed to `onDisconnect()` just like they are passed the resolvers. This way, there would be one consistent way of tracking users/connections.

mwood23:
Here's a [relevant thread on Apollo's Spectrum](https://spectrum.chat/apollo/apollo-server/how-to-access-context-in-ondisconnect-for-subscriptions~81dc7677-1907-4e76-b13b-875d14b8aaf3) that has a code snippet on accessing `onConnect()`'s context in `onDisconnect()`. Appears to work for me!

OmgImAlexis:
Here's a simple version of what's shown in @mwood23's link.

```js
subscriptions: {
    onConnect: connectionParams => {
        const apiKey = connectionParams['x-api-key'];
        ensureApiKey(apiKey);

        const user = Users.findOne({apiKey}) || {name: 'guest'};

        log.info(`<ws> ${user.name} connected.`);

        return {
            user
        };
    },
    onDisconnect: async (_, context) => {
        const initialContext = await context.initPromise;
        log.info(`<ws> ${initialContext.user.name} disconnected.`);
    }
}
```

Issue #94, opened 2017-03-09
Node server keeps crashing repeatedly with "Error: reserved fields must be empty"


booboothefool:
This started happening suddenly, and now I can't even get my server back up. It's been a few hours. It just restarts, then immediately crashes. Using 0.5.1 btw.

Any ideas?

![screen shot 2017-03-09 at 6 52 02 pm](https://cloud.githubusercontent.com/assets/13257645/23776040/b53ac3b0-04f9-11e7-8c47-d6bb5fcd42b8.png)

Very scary...
![screen shot 2017-03-09 at 8 26 49 pm](https://cloud.githubusercontent.com/assets/13257645/23778299/8892e1ea-0507-11e7-9da4-2da9d68c014a.png)

booboothefool:
Where do I add this:

https://github.com/websockets/ws/issues/1039#issuecomment-285591423

For this library?

booboothefool:
I'm trying to add the line here now: https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/server.ts#L96

But installing my fork from GitHub doesn't install the /src directory... huh?

booboothefool:
Alright, so after committing the entire folder to make sure the change was in, which btw was adding:

```
            request.on('error', function (err) {
                console.error(err);
            });
```
to https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/server.ts#L96

the error is being caught, and my server stopped crashing. After about 10 or so 1 star reviews on my app within the past day, hahaha. 

I'm just happy the server doesn't crash, however would like to understand why this is happening if anyone knows. Thanks.

Urigo:
@booboothefool that is a good idea.
Also looks like a pretty simple change, would you like to create a PR to expose that event on the subscriptionServer?

Urigo:
Fixed in https://github.com/apollographql/subscriptions-transport-ws/pull/130

dotansimha:
Thank you for opening this issue, @booboothefool . we fixed it in 0.7.0,
 and now we are handling the `error` callback of the socket,
 as you suggested. 
(https://github.com/apollographql/subscriptions-transport-ws/pull/130)

PR #108, opened 2017-04-13
WIP: Graphql Transport WS Protocol

mistic:
This PR is still a work in progress (with the remain tasks needed to be done described below), that has been done close to @DxCx supervision, but it has already implemented the majority of needed changes to the new package graphql-trasnport-ws according our point 2 on the shared design document ( GraphQL WebSocket transport plan ). So in this PR we have implemented the new protocol in order to support queries, mutations and subscriptions both on server and on the client side (all the code tasks from point 2 of the document are done). 

Please note this is an intermediate pull request to allow you guys start reviewing the most important changes. However a final pull request will be opened from a branch called graphql-transport-ws containing the current code of this branch (graphql-transport-ws-protocol) plus the changes you request in this review, unit tests and those following changes:

- internal code library split into 3 new packages: common, client and server.
- unit tests split into client and server with a common config.

PS: @DxCx thank you so much for your help, support and continuous reviews during the all process!

TODO:

- [x] Fix all the old tests.
- [ ] Code tests to cover all the new logic and  mapping functions for legacy features.
- [x] Add env var to change between hybrid/full ws transport, persistedQueries support, max query size limitation, improve code, and write documentation update for full-transport-ws GitHunt repos. 
- [ ] Update documentation and also including examples for full and hybrid ws transport on it.
- [x] Update CHANGELOG.md with changes made.


Urigo:
@mistic great job!  I love your PR and it helps us move a great step forward.

The direction is exactly as discussed on the design doc and you can move forward.

Here are some comments from reviewing the current state:
 
* I think we should change `GraphQLTransportWSServer` to expose static method for creation without constructor because the instance is useless
* We still want to continue supporting sending queries and mutation over HTTP and subscriptions over Websocket, but now you removed the `addGraphQLSubscriptions` function.  we should keep that functionality, but not here.  it is related to Apollo Client network interface and this package is not tied to Apollo. so please create it as a separate repo and npm package for that and I will add it under the Apollo organization when ready.
* I've tried to run Githunt with your fork but failed with the following error:
screen shot 2017-04-18 at 2 57 53 pm

* Please also include an example of Githunt running with the package both with complete websockets and hybrid with HTTP and Websocket side by side.
* **backward compatibility** Currently the code is not backward compatible because the name changed from `SubscriptionServer` to `GraphQLTransportWSServer` and also `GraphQLTransportWSClient` but, because this would be released as a new npm package, maybe instead of keeping the code back-compack, we can just remove the code and keep the current version on a branch we can support and keep releasing patches from

Great job!

DxCx:
regarding backwards compatibility, we should expose the new names by the old names with deprecation warning.

mistic:
@DxCx I've changed the names both on server and on the client back for the old ones as we discussed.

@Urigo Can u review my updates to see if they matches your expectations? I pointed both Githunt-API and Githunt-react for this code and all the old behaviour with subscriptions over ws and queries and mutations over http is working :D Next I'll test Githunt-API and Githunt-react only over ws with help of @DxCx code.

src/client.ts
```diff
@@ -159,11 +161,11 @@ export class GraphQLTransportWSClient {
     }
 
     if (
-      !isString(query) ||
+      ( !isString(query) && !getOperationAST(query, operationName)) ||
```
DxCx: `query instanceof DocumentNode` is more cleaner.. 👍

src/helpers.ts
```diff
@@ -1,25 +1,36 @@
...
-      }, handler);
+      return wsClient.subscribe(request, handler);
     },
     unsubscribe(id: number): void {
       wsClient.unsubscribe(id);
     },
   });
 }
+
+export function addGraphQLWS(networkInterface: any, wsClientOptions: ClientOptions): any {
```
DxCx: hmm, i might be confused, but why do we need that?

src/helpers.ts
```diff
@@ -1,25 +1,36 @@
...
+  if (typeof networkInterface._uri !== 'string') {
```
DxCx: you are checking if network interface has implementation related uri.. :\

src/client.ts
mistic: It cannot be done in that way :(

mistic:
@DxCx @Urigo I've updated my last comment to reflect how we should use this PR to implement the  hybrid mode and the full ws mode. Is this according your expectations?

src/server.ts
```diff
@@ -160,157 +181,232 @@ export class SubscriptionServer {
...
-        // disconnecting the client 
-        setTimeout(() => {
-          connection.close(1011);
-          connection.terminate();
-        }, 10);
-      }
-    });
+  private sendError(connectionContext: ConnectionContext, reqId: string, errorPayload: any,
+                           overrideDefaultErrorType?: string): void {
+    if ([
```
DxCx: BUG: overrideDefaultErrorType might be undefined.

src/client.ts
```diff
@@ -268,54 +296,73 @@ export class SubscriptionClient {
...
+        this.requests[reqId].handler(this.formatErrors(parsedMessage.payload), null);
+        delete this.requests[reqId];
+        break;
+
+      case MessageTypes.GQL_DATA:
+        const requestPayloadData = parsedMessage.payload.data || null;
+        const requestPayloadErrors = parsedMessage.payload.errors ?
+          this.formatErrors(parsedMessage.payload.errors) : null;
+
+        this.requests[reqId].handler(requestPayloadErrors, requestPayloadData);
```
DxCx: bug: Should be `this.requests[reqId].handler(null, parsedMessage.payload);`

src/client.ts
```diff
@@ -110,38 +97,23 @@ export class SubscriptionClient {
...
-    this.subscriptions[subId] = {options, handler};
-    this.waitingSubscriptions[subId] = true;
-    setTimeout( () => {
-      if (this.waitingSubscriptions[subId]) {
-        handler([new Error('Subscription timed out - no response from server')]);
-        this.unsubscribe(subId);
-      }
-    }, this.subscriptionTimeout);
-    return subId;
+  public subscribe(options: RequestOptions, handler: (error: Error[], result?: any) => void) {
```
DxCx: here we need to add a wrapper handler for converting payload to:
```
const requestPayloadData = parsedMessage.payload.data || null;
const requestPayloadErrors = parsedMessage.payload.errors ?
  this.formatErrors(parsedMessage.payload.errors) : null;
```
to keep backward comptiable with subscription manager

mistic:
@DxCx @Urigo just merged the changes to support executor pattern. I believe at this point I can start writing the unit tests. What do u think guys? There are anything more to do on the features side?

Urigo:
@mistic let's go ahead and finish the tests.
also if you could link to all the relevant Githunt examples we talked about (and add them to your checklist on the opening comment).
Also, please rebase to master since we merged this PR - https://github.com/apollographql/subscriptions-transport-ws/pull/90/

mistic:
@Urigo I'll merge my branch with master right now to integrate with the last changes :D
Regarding to GitHunt can u check my personal messages that I sent you? 

I already finished the GitHunt example integration with the full ws transport. You can check it here:

https://github.com/mistic/GitHunt-API/tree/full-transport-ws
https://github.com/mistic/GitHunt-React/tree/full-ws-transport

However in order to be able to make a PR to GitHunt I need to complete some minor details (like a toggle between hybrid and full ws transport on GitHunt). Let's discuss this.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
+  // send message, or queue it if connection is not open
+  private sendMessageRaw(message: Object) {
+    switch (this.status) {
       case this.client.OPEN:
-        // TODO: throw error if message isn't json serializable?
-        this.client.send(JSON.stringify(message));
+        let serializedMessage: string = JSON.stringify(message);
+        let parsedMessage: any;
+        try {
+          parsedMessage = JSON.parse(serializedMessage);
```
helfer: What is this trying to do? Shouldn't the try/catch be around `JSON.stringify` if at all? What's the benefit of doing a JSON.parse right after JSON.stringify? Right now this looks like an integration test for JSON.stringify and JSON.parse, but I don't think we need to test that, it's pretty well tested. :smile:

If the goal of JSON.parse is to make sure you get back the same thing that you serialized, then I don't think that's something we should do. If people provide objects with prototypes here, that's their problem, not ours.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
         }
     }
   }
 
-  private generateSubscriptionId() {
-    const id = this.maxId;
-    this.maxId += 1;
+  private generateRequestId() {
+    const id = this.nextRequestId;
+    this.nextRequestId += 1;
```
helfer: If we do 10 requests per second, this will overflow after 28 million years. I'm fine with that 🙂

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
 
+  private flushUnsentMessagesQueue() {
+    this.unsentMessagesQueue.forEach((message) => {
+      this.client.send(JSON.stringify(message));
+    });
+    this.unsentMessagesQueue = [];
+  }
+
+  private checkConnection() {
+    !this.wasKeepAliveReceived ? this.close() : this.wasKeepAliveReceived = false;
```
helfer: I don't like ternary statements that start with `!`. It's hard to parse this way. I would just write it with an `if`.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
+
     const delay = this.backoff.duration();
     setTimeout(() => {
       this.connect(true);
     }, delay);
   }
 
+  private flushUnsentMessagesQueue() {
+    this.unsentMessagesQueue.forEach((message) => {
+      this.client.send(JSON.stringify(message));
```
helfer: I think we should pass this through the same serialization logic as when we send messages normally!

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
       return;
     }
 
     if (!this.reconnecting) {
-      this.reconnectSubscriptions = this.subscriptions;
-      this.subscriptions = {};
-      this.waitingSubscriptions = {};
+      Object.keys(this.requests).forEach((key) => {
+        this.unsentMessagesQueue.push(
+          this.buildMessage(parseInt(key, 10), MessageTypes.GQL_START, this.requests[key].options)
```
helfer: Do we have to restart all these operations? How is this handled on the server if there's already an operation with that id running for this client? Are the messages/operation deduplicated there?

src/client.ts
```diff
@@ -268,54 +308,71 @@ export class SubscriptionClient {
...
+    let reqId: number;
+
+    try {
+      parsedMessage = JSON.parse(receivedData);
+      reqId = parsedMessage.id;
+    } catch (e) {
+      throw new Error(`Message must be JSON-parseable. Got: ${receivedData}`);
+    }
+
+    // Complete every request that are not registered on the client but are being sent by the server and are not
```
helfer: I think I really prefer `stop` to `complete`.

src/client.ts
```diff
@@ -87,13 +76,11 @@ export class SubscriptionClient {
     this.connectionParams = connectionParams;
     this.connectionCallback = connectionCallback;
     this.url = url;
-    this.subscriptions = {};
-    this.maxId = 0;
-    this.subscriptionTimeout = timeout;
-    this.waitingSubscriptions = {};
+    this.requests = {};
```
helfer: These should be called `operations`. Request is a one-time thing in most definitions, operation indicates that it can be an ongoing thing. It also makes sense, because query, mutation and subscription are all called "operations" in GraphQL.

src/client.ts
```diff
@@ -268,54 +308,71 @@ export class SubscriptionClient {
...
-        case KEEPALIVE:
-          break;
-
-        default:
-          throw new Error('Invalid message type!');
-      }
-    };
+      case MessageTypes.GQL_CONNECTION_ACK:
+        if (this.connectionCallback) {
+          this.connectionCallback();
```
helfer: I think this is something we were doing for subscriptions, right? Maybe I even implemented this the first time around, but I don't like it that much any more 😬  It's kind of overloading the callback. Do we really need to know when it was connected? I might prefer exposing that in a `connectionStatus()` that can be called from outside, or a `onConnectionStatusChange()` callback. That would feel a lot less hacky and be much more versatile.

src/client.ts
```diff
@@ -268,54 +308,71 @@ export class SubscriptionClient {
...
+    try {
+      parsedMessage = JSON.parse(receivedData);
+      reqId = parsedMessage.id;
+    } catch (e) {
+      throw new Error(`Message must be JSON-parseable. Got: ${receivedData}`);
+    }
+
+    // Complete every request that are not registered on the client but are being sent by the server and are not
+    // in the allowed list
+    if (
```
helfer: This is a bit hard to read. I think if the code was easier to understand, the comment wouldn't be needed.

src/helpers.ts
```diff
@@ -1,14 +1,19 @@
...
+
+/**
+ * @deprecated This method will become deprecated in the new package graphql-transport-ws.
+ * Start using the GraphQLTransportWSClient to make queries, mutations and subscriptions over websockets.
+ */
 
 // Quick way to add the subscribe and unsubscribe functions to the network interface
+// We will move this into a new package in the future
 export function addGraphQLSubscriptions(networkInterface: any, wsClient: SubscriptionClient): any {
+  console.warn('This method becomes deprecated in the new package graphql-transport-ws. Start using the ' +
```
helfer: Let's make sure this warning is not printed if NODE_ENV is production.

src/message-types.ts
```diff
@@ -0,0 +1,47 @@
...
+  public static GQL_CONNECTION_ACK = 'connection_ack';
+  public static GQL_CONNECTION_ERROR = 'connection_error';
+  // NOTE: This one here don't follow the standard due to connection optimization
+  public static GQL_CONNECTION_KEEP_ALIVE = 'ka';
+  public static GQL_CONNECTION_TERMINATE = 'connection_terminate';
+  public static GQL_START = 'start';
+  public static GQL_DATA = 'data';
+  public static GQL_ERROR = 'error';
+  public static GQL_COMPLETE = 'complete';
+  public static GQL_END = 'end';
```
helfer: What is the difference between `COMPLETE` and `END`? If end is to stop an operation, we should call it `STOP`.

src/message-types.ts
```diff
@@ -0,0 +1,47 @@
...
+  // NOTE: Those message types are deprecated and will be removed soon.
```
helfer: In the new package we shouldn't have these at all.

src/helpers.ts
```diff
@@ -1,14 +1,19 @@
 import { SubscriptionClient } from './client';
-import { print } from 'graphql/language/printer';
+
+/**
+ * @deprecated This method will become deprecated in the new package graphql-transport-ws.
+ * Start using the GraphQLTransportWSClient to make queries, mutations and subscriptions over websockets.
+ */
 
 // Quick way to add the subscribe and unsubscribe functions to the network interface
+// We will move this into a new package in the future
```
helfer: In the new package we shouldn't have this, but we can keep it in the old one and just print the deprecation warning.

src/message-types.ts
DxCx: COMPLETE => Means no data from server
END => Means client doesn't want anymore data.

src/client.ts
```diff
@@ -110,38 +97,35 @@ export class SubscriptionClient {
     this.client.close();
   }
 
-  public subscribe(options: SubscriptionOptions, handler: (error: Error[], result?: any) => void) {
-    const { query, variables, operationName, context } = options;
-
-    if (!query) {
-      throw new Error('Must provide `query` to subscribe.');
-    }
+  public query(options: RequestOptions): Promise<ExecutionResult> {
```
DxCx: well @helfer at the moment we will be compatible with current apollo,
however, it is built in a way that when you will go forward to use 1 method that returns an Observable, the changes here are minimal.

src/client.ts
DxCx: this is due to the old comment `// TODO: throw error if message isn't json serializable?` which was written by you.. XD

src/client.ts
DxCx: it will tear down old request and start a new one instead

src/message-types.ts
DxCx: that's exactly what this comment means.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
-      this.unsubscribe(parseInt(subId));
+    Object.keys(this.requests).forEach( subId => {
+      this.unsubscribe(parseInt(subId, 10));
     });
   }
 
-  // send message, or queue it if connection is not open
-  private sendMessage(message: Object) {
-    switch (this.client.readyState) {
+  private executeRequest(options: RequestOptions, handler: (error: Error[], result?: any) => void): number {
```
helfer: Actually, I think I'm fine with `operationOptions`. But another point here. If all operations can be long running, we need a way to cancel them, so `executeRequest` could for example return a function, which when called will stop the operation, just like it works with observables.

src/client.ts
helfer: Ok, I think if @Urigo prefers it this way, it's fine 🙂  We can rewrite it later.

src/message-types.ts
helfer: Ok, let's call that STOP. end is more commonly used as a noun. Stop is almost always a verb.

src/client.ts
helfer: No need to change this right now.

src/client.ts
helfer: Ok, then I think the test should be around `JSON.stringify`. No need to re-parse.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
+    this.sendMessage(reqId, MessageTypes.GQL_START, options);
+
+    return reqId;
+  }
+
+  private buildMessage(id: number, type: string, payload: any) {
+    const payloadToReturn = payload && payload.query ?
+      {
+        ...payload,
+        query: typeof payload.query === 'string' ? payload.query : print(payload.query),
```
helfer: No need to change it right now, it's ok to keep it for backcompat.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
+    if (!query) {
+      throw new Error('Must provide a `query`.');
+    }
+
+    if (!handler) {
+      throw new Error('Must provide an `handler`.');
+    }
+
+    if (
+      ( !isString(query) && !getOperationAST(query, operationName)) ||
```
helfer: No need to change it right now, it's ok to keep it for backcompat.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
-    delete this.waitingSubscriptions[id];
-    let message = { id: id, type: SUBSCRIPTION_END};
-    this.sendMessage(message);
+    if (this.requests[id]) {
+      delete this.requests[id];
+    }
+    this.sendMessage(id, MessageTypes.GQL_END, undefined);
   }
 
   public unsubscribeAll() {
```
helfer: No need to change it right now, it's ok to keep it for backcompat.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
   }
 
   public unsubscribe(id: number) {
```
helfer: No need to change it right now, it's ok to keep it for backcompat.

src/client.ts
```diff
@@ -110,38 +97,35 @@ export class SubscriptionClient {
...
 
-    if (
-      !isString(query) ||
-      ( operationName && !isString(operationName)) ||
-      ( variables && !isObject(variables))
-    ) {
-      throw new Error('Incorrect option types to subscribe. `subscription` must be a string,' +
-      '`operationName` must be a string, and `variables` must be an object.');
-    }
+  public subscribe(options: RequestOptions, handler: (error: Error[], result?: any) => void) {
```
helfer: No need to change it right now, it's ok to keep it for backcompat.

src/server.ts
```diff
@@ -158,156 +336,227 @@ export class SubscriptionServer {
...
-
-    connection.send(JSON.stringify(message));
-  }
+  private sendError(connectionContext: ConnectionContext, reqId: string, errorPayload: any,
+                           overrideDefaultErrorType?: string): void {
+    const sanitizedOverrideDefaultErrorType = overrideDefaultErrorType || MessageTypes.GQL_ERROR;
+    if ([
+        MessageTypes.GQL_CONNECTION_ERROR,
+        MessageTypes.GQL_ERROR,
+      ].indexOf(overrideDefaultErrorType) === -1) {
```
DxCx: you forgot to use `sanitizedOverrideDefaultErrorType` in here.

mistic: Yeah thank you :D

src/message-types.ts
mistic: Done

src/message-types.ts
mistic: As discussed let's keep things related with backwards compatibility for now.

src/helpers.ts
mistic: As discussed let's keep things related with backwards compatibility for now.

src/client.ts
mistic: I think you already get the difference between complete and stop :)

src/client.ts
mistic: So lets keep this one as is

src/client.ts
mistic: As discussed let's keep things related with backwards compatibility for now.

src/client.ts
mistic: As discussed let's keep things related with backwards compatibility for now.

src/client.ts
mistic: As discussed let's keep things related with backwards compatibility for now.

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
   }
 
-  // send message, or queue it if connection is not open
-  private sendMessage(message: Object) {
-    switch (this.client.readyState) {
+  private executeRequest(options: RequestOptions, handler: (error: Error[], result?: any) => void): number {
+    const { query, variables, operationName } = options;
+
+    if (!query) {
+      throw new Error('Must provide a `query`.');
```
mistic: Done

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
+    }
+
+    if (!handler) {
+      throw new Error('Must provide an `handler`.');
```
mistic: Done regarding the backticks. However I think we should right "an hander". The "h" is a special case here.

src/client.ts
mistic: we only get an exception return from JSON.parse not from JSON.stringify

src/client.ts
mistic: done

src/client.ts
mistic: done

src/client.ts
mistic: Done

src/client.ts
```diff
@@ -1,42 +1,33 @@
...
-  INIT_FAIL,
-  INIT_SUCCESS,
-} from './messageTypes';
-import { GRAPHQL_SUBSCRIPTIONS } from './protocols';
-
+import * as Backoff from 'backo2';
+import { EventEmitter, ListenerFn } from 'eventemitter3';
 import isString = require('lodash.isstring');
 import isObject = require('lodash.isobject');
+import { ExecutionResult, print, getOperationAST } from 'graphql';
```
mistic: Done

src/client.ts
```diff
@@ -165,95 +149,151 @@ export class SubscriptionClient {
...
 
+  // ensure we have an array of errors
+  private formatErrors(errors: any) {
+    if (Array.isArray(errors)) {
+      return errors;
+    }
+    if (errors && errors.message) {
+      return [errors];
+    }
+    return [{ message: 'Unknown error' }];
```
mistic: done

src/client.ts
NeoPhi: `JSON.stringify` can throw errors due to circular structures. `const a = {}; const b ={a}; a.b = b; JSON.stringify(a);`

src/client.ts
mistic: done

dotansimha:
@mistic @DxCx @Urigo 
Noticed an issue earlier: In some cases, the client disconnects, but the onClose of the client isn't called, and the Network tab of the browser still thinks the client is connected. 
The server side tried to publish data to this socket and getting "Not opened" error (the connection state is 3 = CLOSED).

mistic:
@dotansimha Have u tested with last ws version? Ins't the problem related with ws library itself? 
https://github.com/websockets/ws/issues/686

mistic:
Guys ( @DxCx @helfer @NeoPhi @Urigo @dotansimha or anyone else interested in give some help ) will be really appreciated some help with those 2 tasks:

- Update new documentation.
- Update new CHANGELOG.md with changes made.

helfer:
@mistic wow, great work! I've been really busy myself, so I couldn't help or even review until now. If Uri can't review, I hope I can do it over the weekend.

mxstbr:
Do I understand correctly that this PR makes it possible to run the GraphQL and the subscriptions server off of the same server and thus the same port? (which is a requirement for deploying on now.sh)

I've tried taking the code sample above and using the branch this PR lives in, but I can't seem to make it work. Are there any docs yet so I can try this out and give some feedback?

mistic:
@mxstbr You'll be able to understand how to use this PR looking for those already changed GitHunt example to support this PR:

https://github.com/mistic/GitHunt-API/tree/full-transport-ws
https://github.com/mistic/GitHunt-React/tree/full-ws-transport

mistic:
Guys ( @DxCx @helfer @NeoPhi @Urigo @dotansimha ) I think we're reaching the end of the work on this PR. Today I merged into this PR the Uri changes to correctly support subscribe method from `graphql-js` into our server  graphql executor. Also I already wrote the new `CHANGELOG.md`.

Work needed to be done to merge this PR:

- Write some tests to cover the new logic and the mapping features. All the old ones are already updated and being pass.
- Update documentation to reflect the new API on server and client and also provide examples.

Urigo:
@mistic amazing amazing work.
I can't thank you enough for all your effort.
There are still more tasks to be done, but I'm going to merge this now because I want @thebigredgeek and @srtucker22 to be able to work and push their changes on top of the new API.

During the next couple of days:

1.   Add tests for the new API - @mistic
2.   Add documentation for the new API - @mistic
3.  @thebigredgeek and @srtucker22 - Rebase your PR against the latest master and the new API.
4.  I will make sure to release a version `0.7.0` that supports the current API without breaking changes, but also include the new API based on the upcoming `graphql-js` with subscriptions
5. I will remove all the old API code and will release a new package called `graphql-ws-transport`

Urigo:
@mistic , thank you again for your great PR.
During your work, a new `graphql.js` [subscriptions support came to be](https://github.com/graphql/graphql-js/pull/846).

This changes a bit our goals so I want to reflect on that for a sec, to create an updated path with slight adjustments.

**V0.7 goals**:

- Full backward compatibility
- New API
   - Support query and mutations over WebSocket, with option for hybrid.
   - Support new graphql.js subscriptions with AsyncIterable

**Changes by @mistic** :

- Added support for full websocket network interface.
- Added support for hybrid network interface (using `addGraphQLSubscriptions` method)
- Implemented new transport protocol
- Added `ExecuteAdapters` to support both old API (`SubscriptionsManager`) and new API (graphql `subscribe`)
- Added `executeReactive` as default payload source (no need to run `ExecuteAdapter`), and implemented the transport with Observable base. 

**On top of those changes, I’ve added the support for the new graphql.js subscriptions implementation**:
- Transport work natively with AsyncIterable for payload source (but requires `ExecuteAdapters`)
- Added ability for the transport to expect `execute` and `subscribe` directly from `graphql-js` new subscription implementation. 
   - **Tiago changed the API from `new SubscriptionsClient({  execute, subscribe   })` to `new SubscriptionsClient({  executer: { execute, subscribe }   })` in order to make `executeReactive` as named priority**
- Transport works with the recent `graphql-subscriptions` master (with both new and old API).

**The problems I have with the current state are those**:
- We understood from the graphql.js subscriptions work that in most cases, we should separate initiating the subscriptions and the firing events [Lee’s comment here](https://github.com/graphql/graphql-js/pull/846#issuecomment-300935894)
- That means we want to make that common API as simple as possible and the default one instead of the special `executeReactive` API
- Once we do that, we should come up with a better design on how to support the `executeReactive` approach without complexing the common API and implementation

**So the changes I'm proposing to do before releasing 0.7 are**:
- Transport work natively with `AsyncIterable` for payload source to make the most common use case the simplest and the default one supported (instead of `executeReactive`)
- Transport should expect `execute` and `subscribe` directly from `graphql-js` new subscription implementation 
- bring back the simple ` new SubscriptionsClient({  execute, subscribe   })` API
- In order to support `executeReactive` while keeping this library simple, I believe the right solution should be to add a different executor library for that approach. That library should use this transport like so: expose `subscribe` and `execute` directly and include an adapter for Observable -> AsyncIterable internally, instead of placing that code inside this library
- Merge the two open PRs to help with authentication:
   - https://github.com/apollographql/subscriptions-transport-ws/pull/124
   - https://github.com/apollographql/subscriptions-transport-ws/pull/78
- Solve those issues:
   - https://github.com/apollographql/subscriptions-transport-ws/issues/99
   - https://github.com/apollographql/subscriptions-transport-ws/issues/95
   - https://github.com/apollographql/subscriptions-transport-ws/issues/94
- Tests for all the new changes and APIs
- Docs for all the new changes and APIs

**V0.8 goals**:
- Clean and remove all backward compatible code

**V0.9 goals**:
- Ability to replace `ws` library with `uws` or any other WS implementation library

@DxCx @mistic @helfer @stubailo I want to hear your opinions on that plan

DxCx:
@Urigo the work on Websocket is done for ~0.5year now,
I've already written it 3 times, while Tiago now did the 4th time with me helping him,

we used observables back then and that's the reason it has observables, AsyncIterator was not in the picture until last month. (more then that, supporting Observables in GraphiQL was already in progress, which made sense why using it).

if you want to change executeReactive signature to return AsyncIterator, I'm totally fine with that because graphql internally started to use, but we shouldn't remove executeReactive, if you have a better proposed solution, let me know.

so, unless you know that plans for how facebook is going to support reactive directives such as `@defer` or `@live` we need a reactive execute function to be supported,
and the `executor` pattern is great because it gives you great flexibility.

about the versioning, generally i agree about the content but i think that V0.8 Should be V1.0 or something like that because it will have a lot of breaking changes.

Urigo:
@DxCx 

1. I totally understand you pain, but let's try to solve this together. that's why I wrote this comment.
2. Let's separate WebSocket work and @live. WebSocket is now merged and it can work without `executeReactive`. You now will be able to send queries and mutations over WS as well which is something a lot of people as asked for. So it doesn't matter how long WebSocket work has been done, now it's in master and it can work even without `executeReactive`.
3. `AsyncIterator` wasn't a thing, that's why I'm writing this comment, because now it is a thing.  If we really want to build the tools that help GraphQL developers use the main reference implementation and help them do that in production environment with the minimum changes, we should change that API as little as possible. On other libraries we can create completely separate APIs, but those are different libraries with different goals.  We should be really careful about what is the goal for each library.
4. I want to work together in order for you to be able to use as much from this library as possible in your solution, without enforcing everyone to add size and complexity if they are not going to use your solution. I would love to hear more design ideas from you about how to do that. Do you think that there is a solution where you wrap the API that I described in another library and exposing the API you want without hurting the API that 90% of the users will use?

I agree with you about versioning. let's do a `v1.0` after `v0.7`.

DxCx:
alright, @Urigo so as i suggested we can move implementation to async iterator instead of observable, and that will be the thinest possible, i don't understand why do we need to completely remove `executeReactive` support if it won't really hurt anyone
+ it will let people with legacy subscription server upgrade to run mutation + queries with one or two lines of change, not needing to rebuild their whole schemas.

mistic:
@Urigo I agree with @DxCx in almost everything! 

1 - the v0.8 should be the v1.0.0. It will have a lot of breaking changes and it's almost a completely rewrite. In my opinion is a major version. 

2 - Regarding to the SubscriptionsServer, what's the problem with the executor pattern? In my opinion this isn't a deal breaker with anything coming from graphql-js. We just accept an executor in our Server that can be composed by `execute & subscribe` from graphql-js, by `executeReactive` from graphql-subscriptions or simply `execute` from graphql-subscriptions and `subscribe` from graphql-js. What's the real problem with this thing? Is simply a design choice from our server.  In the limit we could provide some predefined executor interfaces to provide more context to the user when instantiating the server.

3 - If you prefer we can move observables to AsyncIterator.

dotansimha:
@mistic @DxCx @Urigo 
What do you think about accepting either `Promise` or `AsyncIterator` for `execute` function? 

This means that Hagai can implement `@live` logic inside a separate package, and expose two functions: `execute` and `subscribe` (just like `graphql-js` does).

The transport can check if the return value is `AsyncIterator` and wait for reactive changes and publish it over the socket for each value, or use `Promise` and publish it only once.

DxCx:
@dotansimha that works as well!
my only problem with that is we did not get Lee's confirmation.
which means that we are going to modify a signature of an existing function.
so, if future plans of facebook is to modify execute signature, that's fine.
but if not, it's better to just introduce a new one IMO.
which is essentially the `executeReactive` introduced here.
as both me & @mistic stated above, there is no problem to eliminate observable reference and support only AsyncIterator

dotansimha:
@DxCx 
We don't need to modify the `execute` signature in GraphQL. We just need to accept a return value of both options, and handle it correctly inside this transport.
It does not require any modifications in `graphql-js`, the changes could be in here:
https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/server.ts#L53-L60

And change it to:
```ts
export type ExecuteFunction = (
  schema: GraphQLSchema,
  document: DocumentNode,
  rootValue?: any,
  contextValue?: any,
  variableValues?: {[key: string]: any},
  operationName?: string,
) => Promise<ExecutionResult> | AsyncIterator<ExecutionResult>;
```

This way, `graphql-js` remains the same, but the transport can support both `Promise` and `AsyncIterable`.

What do you think?

DxCx:
@dotansimha i totally understood your meaning, as i said it could work.
but then, don't matter which package will be able to return AsyncIterator as execute, it won't be compatible with other graphql tools except for this package.
the only reason to do it is if reference implementation are going towards this direction as well instead of introducing a new call, but i really don't believe this will happen as it is a breaking change.

dotansimha:
@DxCx 
What about exposing `AsyncIterator` along with `IObservable` from packages? 

For example, Hagai's `ReacticeExecutor` can expose `execute` that returns `AsyncIterator` along with `executeObservable` that returns `IObservable`.

DxCx:
@dotansimha as i mentioned above, i don't mind terminating the observable.
the only intention with observables is to be able to return reactive result. `AsyncIterator` gives exactly that as well.
the idea is to have an execute signature that also returns an AsyncIterator, if you suggested executeObservable then i really don't see a reason why not keep the existing name of `executeReactive` which returns `AsyncIterator` instead of `Observable`?

dotansimha:
I meant to expose `executeObservable` from your GraphQL alternative executer package (the implementation that in your `graphlq-subscription` PR)

DxCx:
@dotansimha i think there is a confusion in here, this is what i'm trying to achieve:
1. change exiting executeReactive signature to return AsyncIterator instead of observable
2. change all inner observable implementation to AsyncIterator, removing Observable code completely
3. change GraphQLExecutorWithSubscription (graphql-subscriptions) to return AsyncIterator and not observable

Do you agree?

dotansimha:
@DxCx Yeah that's sound great.

Also, I think that since we can remove `SubscriptionsManager` in a future version, and the dependency for `graphql-js` in `graphql-subscritpions` - maybe it's better to implement [this](https://github.com/apollographql/graphql-subscriptions/pull/47) in a new package?

I think that the future version of `graphql-subscriptions` will only contain utils like channel mapping, filters and such.

What do you think?

DxCx:
the idea behaind it was to allow people people that are using `graphql-subscriptions` today to upgrade easily to full graphql over websocket.
there is no reason for that patch as well if there is no SubscriptionManager.

Issue #123, opened 2017-05-10
lazy connect / connectionParams


thebigredgeek:
Auth seems to be a big sticking point, as mentioned in several issues:

https://github.com/apollographql/subscriptions-transport-ws/issues/33
https://github.com/apollographql/subscriptions-transport-ws/issues/75

I am starting to encounter pain with auth, myself.  I am building a react-native app where I need to be able to connect the WS only after the user is authenticated to protect the WS server.

Right now, on our web app, I am passing a JWT token (if available) into `connectionParams` and then parsing that token server side... rejecting the connection if the token isn't present or is invalid or belongs to a user who should no longer be able to access our product.  This works well for our current build, because we set a cookie and do full page refresh after and so we get a "fresh" network interface that has the ws transport applied to it whenever the token cookie is present.

However, with react native, I can't simply "refresh", nor do I really want to as I want to experience to feel fluid as most mobile apps feel.

I need a way to conditionally connect with a dynamic connection param at the time that the user finishes logging in or registering.

My proposal is that we should modify `connectionParams` to accept either an object (as is) or a function that returns an object.  If `connectionParams` is a function, the `connect` call in the `SubscriptionClient` constructor is skipped and lazily invoked on the first `subscribe` call.  This would satisfy my case, because I don't reach code that attempts subscriptions until the user is in a state of authentication.

I would like to submit a PR for this tomorrow.  This is a huge blocker for me right now.

Thanks!

hmlb:
Thank you for this work @thebigredgeek. I have exactly the same need as you (but for angular-apollo client).

`connectionParams` as a function is a nice way of solving this. When your credentials are updated, we would need to call `close` on `SubscriptionClient`. It will reconnect automatically with the new credentials from `connectionParams` and the context will be updated on the server side only for the connections initialized with `reconnect: true`. It would not be an issue for me but do you think it could be ?

thebigredgeek:
@hmlb for my own use case, I don't need a WS connection until after the user has been authenticated... and this is probably best practice in terms of WS security.  The `lazy` flag causes the interface to not attempt a WS connection until the first `subscribe` call is made, which means there is no need to `close`/`reconnect`.

hmlb:
That's fine.  I now use your fork of `SubscriptionClient` and it works perfectly for the lazy (post login) first connection, and for my more specific use case of different levels of authorization where I need to `close`/`reconnect` for the same user.
Thanks !

[...]

PR #124, opened 2017-05-10
lazy mode and deferred connection params

thebigredgeek:
TODO:


thebigredgeek:
Will write tests once the design is approved.

thebigredgeek:
https://github.com/apollographql/subscriptions-transport-ws/issues/123

thebigredgeek:
bump @helfer @Urigo @stubailo

helfer:
@thebigredgeek I think @Urigo (busy atm) or @NeoPhi should review this. Reviewing will be a lot easier and PRs will be merged faster if you make this two separate (maybe stacked) PRs, one for the connectionParams function, one for the lazy connecting. I think connectionParams function is pretty easy to test and merge, lazy connecting might require more testing and more discussions.

thebigredgeek:
I can separate them, but merging one but not the other doesn't help me unfortunately.  I'll try to find time for that if it ends up being necessary

Urigo:
@thebigredgeek great job.
I like both the `connectionParams` and the lazy connecting.
Few comments:

1. I don't see a reason to explicitly convert to `Boolean`. we never use it so can we remove that?
2. Rebase against the latest master and make sure it work with the new API introduced [here](https://github.com/apollographql/subscriptions-transport-ws/pull/108)
3. @srtucker22 I think this touches on the same problem that you solved on [your PR](https://github.com/apollographql/subscriptions-transport-ws/pull/78). Your PR though creates a more robust approach with middlewares.  do you think we should introduce both solutions or just use one of those approaches for simplicity of the API? also @thebigredgeek let me know your thoughts on that.

srtucker22:
I personally like both approaches :) 
@thebigredgeek made a good point that he preferred to only have authenticated WS connections to avoid needing to build in a lot of DOS security. There are also occasions where ppl will require an open connection without auth ala Meteor and would prefer to just keep the connection open instead of disconnect/reconnect when auth changes. They'll just have to employ some extra security (as you need to do with meteor) to secure against DOS when ppl load up on unauthed connections. 

I also think there is a place for middleware before a subscription for other use cases. Aaand I personally like authorizing all my subscription requests for added security instead of relying on just the connection params.

So I think both pieces are useful separately and together. That's just my $.02

thebigredgeek:
Yeah I agree.  I think the middleware stuff is super useful, but I want to be able to reject connections as they occur rather than just relying on ACL to determine whether a user should be able to subscribe etc.  My main use case here is react native... where I can't do a "full page refresh" after login (which in my web app, login state is used to determine whether or not to decorate with the subscription transport network interface).

thebigredgeek:
@Urigo 
1) Will do, I just tend to code more defensively with respect boolean stuff due to PTSD from earlier versions of JS.  Old habits die hard, but should probably die at this point ;)
2) Yup, will do in the AM :)

thebigredgeek:
Also @Urigo how do we want to test this?  Seems like it might break a lot of the existing tests

Urigo:
@thebigredgeek can you please explain why it might break the existing tests? the default behavior should be with `lazy` flag set to `false`, and the `connectionParams` as regular object, so the existing tests should be fine.

Urigo:
@thebigredgeek I've also merged @srtucker22 changes, so please rebase again

thebigredgeek:
@Urigo ah yeah true.  I guess the defaults to existing behavior means no breaking tests :D.  Cool, I'll rebase and add test coverage.

Urigo:
thank you very much @thebigredgeek !

PR #134, opened 2017-05-18
API Docs

dotansimha:
@Urigo @mistic  @DxCx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,13 @@
 ### vNEXT
 - Client exposes new asyncronous middleware to modify `OperationOptions` [PR #78](https://github.com/apollographql/subscriptions-transport-ws/pull/78)
 - Added `WebSocketServer` error handler to prevent uncaught exceptions. Fixes [Issue #94](https://github.com/apollographql/subscriptions-transport-ws/issues/94)
+- Updated `ws` dependency to the lastest.
 - Introduce lazy mode for connection, and accept function as `connectionParams` [PR #131](https://github.com/apollographql/subscriptions-transport-ws/pull/131)
 - Extend transport protocol to support GraphQL queries and mutations over WebSocket [PR #108](https://github.com/apollographql/subscriptions-transport-ws/pull/108) 
 - Added built-in support for `subscribe` from `graphql-js` [PR #133](https://github.com/apollographql/subscriptions-transport-ws/pull/133)
+- Fixed infinity reconnects when server accepts connections but its in an error state. [PR #135](https://github.com/apollographql/subscriptions-transport-ws/pull/135)
+- Force close client-side socket when using `close()`, and ignore reconnect logic. [PR #137](https://github.com/apollographql/subscriptions-transport-ws/pull/137)
+
 
 ### 0.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ### vNEXT
 - Client exposes new asyncronous middleware to modify `OperationOptions` [PR #78](https://github.com/apollographql/subscriptions-transport-ws/pull/78)
 - Added `WebSocketServer` error handler to prevent uncaught exceptions. Fixes [Issue #94](https://github.com/apollographql/subscriptions-transport-ws/issues/94)
-- Introduce lazy mode for connection, and accept function as `connectionParams` [PR #131]
+- Introduce lazy mode for connection, and accept function as `connectionParams` [PR #131](https://github.com/apollographql/subscriptions-transport-ws/pull/131)
+- Extend transport protocol to support GraphQL queries and mutations over WebSocket [PR #108](https://github.com/apollographql/subscriptions-transport-ws/pull/108) 
+- Added built-in support for `subscribe` from `graphql-js` [PR #133](https://github.com/apollographql/subscriptions-transport-ws/pull/133)
 
 ### 0.6.0
 
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -21,6 +21,8 @@ export interface OperationMessage {
 #### GQL_CONNECTION_INIT
 Client sends this message after plain websocket connection to start the communication with the server
 
+The server will response only with `GQL_CONNECTION_ACK` or `GQL_CONNECTION_ERROR` to this message.
+
 - `payload: Object` : optional parameters that the client specifies in `connectionParams`
 
 #### GQL_START
@@ -50,7 +52,6 @@ It server also respond with this message in case of a parsing errors of the mess
 #### GQL_CONNECTION_ACK
 The server may responses with this message to the `GQL_CONNECTION_INIT` from client, indicates the server accepted the connection.
 
-The server will response only with `GQL_CONNECTION_ACK` or `GQL_CONNECTION_ERROR` to `GQL_CONNECTION_INIT` message.
 
 #### GQL_DATA
 The server sends this message to transfter the GraphQL execution result from the server to the client, this message is a response for `GQL_START` message.
@@ -68,7 +69,8 @@ Server sends this message upon a failing operation, before the GraphQL execution
 - `id: string` : operation ID of the operation that failed on the server
 
 #### GQL_COMPLETE
-Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive (for queries and mutations, it called immediatly, and for subscriptions it called when unsubscribing)
+Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive for the specific operation.
+
 - `id: string` : operation ID of the operation that completed
 
 #### GQL_CONNECTION_KEEP_ALIVE
@@ -79,10 +81,6 @@ The client starts to considerer the keep alive message only upon the first recei
 ### Messages Flow
 
 This is a demonstration of client-server communication, in order to get a better understanding of the protocol flow:
-
-
-- Server configured with GraphQL schema (and `PubSub` implementation when using subscriptions)
-- Client creates a WebSocket instance using `SubscriptionsClient` object.
 
 #### Session Init Phase
 
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -4,7 +4,10 @@
 
 Each message has a `type` field, which defined in the protocol of this package, as well as associated fields inside `payload` field, depending on the message type, and `id` field so the client can identify each response from the server.
 
-The structure of each client side message is:
+Each WebSocket message is represented in JSON structure, and being stringified before sending it over the network.
+
+This is the structure of each message:
+
 ```typescript
 export interface OperationMessage {
   payload?: any;
@@ -18,12 +21,7 @@ export interface OperationMessage {
 #### GQL_CONNECTION_INIT
 Client sends this message after plain websocket connection to start the communication with the server
 
-This message also triggers the `onConnect` callback on the server.
-
 - `payload: Object` : optional parameters that the client specifies in `connectionParams`
-
-#### GQL_CONNECTION_TERMINATE
-Client sends this message to terminate the connection.
 
 #### GQL_START
 Client sends this message to execute GraphQL operation
@@ -32,38 +30,45 @@ Client sends this message to execute GraphQL operation
     * `query: string` : GraphQL operation as string or parsed GraphQL document node
     * `variables?: Object` : Object with GraphQL variables
     * `operationName?: string` : GraphQL operation name
-    * `context?: any` : Execution context for the operation
     
 #### GQL_STOP
 Client sends this message in order to stop a running GraphQL operation execution (for example: unsubscribe)
 - `id: string` : operation id
     
+#### GQL_CONNECTION_TERMINATE
+Client sends this message to terminate the connection.    
+    
 ### Server -> Client
 
 #### GQL_CONNECTION_ERROR
-The server sends this message if `onConnect` callback returns `false` or throws an exception, and after sending this message, the server disconnects the client.
+The server may responses with this message to the `GQL_CONNECTION_INIT` from client, indicates the server rejected the connection.
 
 It server also respond with this message in case of a parsing errors of the message (which does not disconnect the client, just ignore the message).
 
 - `payload: Object`: the server side error
 
 #### GQL_CONNECTION_ACK
-The server sends this message if `onConnect` callback returns any other value then `false`.
+The server may responses with this message to the `GQL_CONNECTION_INIT` from client, indicates the server accepted the connection.
+
+The server will response only with `GQL_CONNECTION_ACK` or `GQL_CONNECTION_ERROR` to `GQL_CONNECTION_INIT` message.
 
 #### GQL_DATA
-The server sends this message to transfter the GraphQL execution result from the server to the client.
+The server sends this message to transfter the GraphQL execution result from the server to the client, this message is a response for `GQL_START` message.
+
+For each GraphQL operation send with `GQL_START`, the server will respond with at least one `GQL_DATA` message.
+
 - `id: string` : ID of the operation that was successfully set up
 - `payload: Object` : 
     * `data: any`: Execution result
     * `errors?: Error[]` : Array of resolvers errors
 
 #### GQL_ERROR
-Server sends this message upon failing to **register** or **execute** a GraphQL operation (resolver errors are part of `GQL_DATA` message, and will be added as `errors` array)
-- `payload: { errors: Array<Object> }` : payload with an array of errors attributed to the subscription failing on the server
+Server sends this message upon a failing operation, before the GraphQL execution, usually due to GraphQL validation errors (resolver errors are part of `GQL_DATA` message, and will be added as `errors` array)
+- `payload: Error` : payload with the error attributed to the operation failing on the server
 - `id: string` : operation ID of the operation that failed on the server
 
 #### GQL_COMPLETE
-Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive (for queries and subscription, it called immediatly, and for subscriptions it called when unsubscribing)
+Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive (for queries and mutations, it called immediatly, and for subscriptions it called when unsubscribing)
 - `id: string` : operation ID of the operation that completed
 
 #### GQL_CONNECTION_KEEP_ALIVE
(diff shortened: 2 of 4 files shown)


PROTOCOL.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### GQL_CONNECTION_INIT
+Client sends this message after connecting, this triggers `onConnect` on the server.
+- `payload: Object` : optional parameters that the client specifies in `connectionParams`
+
+#### GQL_CONNECTION_TERMINATE
+Client sends this message to terminate the connection.
+
+#### GQL_START
+Client sends this message to execute GraphQL operation
```
mistic: This should be:
- id: string: The id of the GraphQL operation to start
- payload: Object
  -- `query: string` : GraphQL operation as string or parsed GraphQL document node
  -- `variables?: Object` : Object with GraphQL variables
  -- `operationName?: string` : GraphQL operation name
  -- `context?: any` : Execution context for the operation

PROTOCOL.md
```diff
@@ -0,0 +1,74 @@
...
+- `context?: any` : Execution context for the operation
+    
+#### GQL_STOP
+Client sends this message in order to stop a running GraphQL operation execution (for example: unsubscribe)
+- `id: string` : operation id
+    
+### Server -> Client
+
+#### GQL_CONNECTION_ERROR
+The server sends this message if `onConnect` callback returns `false` or throws an exception. after sending this message, the server disconnects the client.
```
mistic: This is not always true. The server always send this `GQL_CONNECTION_ERROR` when it receives a message that is unable to parse, and in this situation it don't disconnects the client

PROTOCOL.md
```diff
@@ -0,0 +1,74 @@
...
+- `payload: Object`: the server side error
+
+#### GQL_CONNECTION_ACK
+The server sends this message if `onConnect` callback returns any other value then `false`.
+
+#### GQL_DATA
+The server sends this message to transfter the GraphQL execution result from the server to the client.
+- `id: string` : ID of the operation that was successfully set up
+- `payload: { data: any }` : Execution result
```
mistic: Why not present this as:

- `id: string`: ...
- `payload: object`
  -- `data: any`: ...
  -- `errors?: []` : ...

PROTOCOL.md
```diff
@@ -0,0 +1,74 @@
...
+Server sends this message upon failing to **register** or **execute** a GraphQL operation (resolver errors are part of `GQL_DATA` message, and will be added as `errors` array)
+- `payload: { errors: Array<Object> }` : payload with an array of errors attributed to the subscription failing on the server
+- `id: string` : operation ID of the operation that failed on the server
+
+#### GQL_COMPLETE
+Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive (for queries and subscription, it called immediatly, and for subscriptions it called when unsubscribing)
+- `id: string` : operation ID of the operation that completed
+
+#### GQL_CONNECTION_KEEP_ALIVE
+Server message sent periodically to keep the client connection alive.
```
mistic: It would be interesting mention here that the client starts to considerer the keep live only upon the first received keep alive message from the server.

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 ### vNEXT
 - Client exposes new asyncronous middleware to modify `OperationOptions` [PR #78](https://github.com/apollographql/subscriptions-transport-ws/pull/78)
```
mistic: It will be good to also mention those PR's here:

https://github.com/apollographql/subscriptions-transport-ws/pull/135
https://github.com/apollographql/subscriptions-transport-ws/pull/137

I believe they are very important and also will be merged. I'm also preparing a third PR to give some refactoring in the changes introduced by PR #78

README.md
```diff
@@ -4,45 +4,43 @@
 
 **(Work in progress!)**
 
-A GraphQL WebSocket server and client to facilitate GraphQL subscriptions.
+A GraphQL WebSocket server and client to facilitate GraphQL queries, mutations and subscriptions over WebSocket.
 
 > `subscriptions-transport-ws` is an extension for GraphQL, and you can use it with any GraphQL client and server (not only Apollo).
 
 See [GitHunt-API](https://github.com/apollostack/GitHunt-API) and [GitHunt-React](https://github.com/apollostack/GitHunt-React) for an example server and client integration.
```
mistic: Just to remember we should also release the new versions of the examples 😄

README.md
```diff
@@ -180,25 +210,27 @@ ReactDOM.render(
...
 
 ## SubscriptionClient
 ### `Constructor(url, options, connectionCallback)`
-- `url: string` : url that the client will connect to
+- `url: string` : url that the client will connect to, starts with `ws://` or `wss://`
 - `options?: Object` : optional, object to modify default client behavior
   * `timeout?: number` : how long the client should wait in ms for a subscription to be started (default 5000 ms)how long the client should wait in ms for a subscription to be started (default 5000 ms)
-  * `connectionParams?: Object` : object that will be available as first argument of `onConnect` (in server side)
+  * `lazy?: boolean` : use to set lazy mode - connects only when first subscription created, and delay the socket initialization
+  * `connectionParams?: Object | Function` : object or function that will be available as first argument of `onConnect` (in server side)
```
mistic: Is this true? Isn't only the connectionParams function used to return the the connectionParams to be sent to the server? The description as is gives the idea that the function is sent 😄

README.md
```diff
@@ -180,25 +210,27 @@ ReactDOM.render(
...
   * `reconnect?: boolean` : automatic reconnect in case of connection error
   * `reconnectionAttempts?: number` : how much reconnect attempts
   * `connectionCallback?: (error) => {}` : optional, callback that called after the first init message, with the error (if there is one)
-- `webSocketImpl: Object` - optional, WebSocket implementation. use this when your environment does not have a built-in native WebSocket (for example, with NodeJS client)
+- `webSocketImpl?: Object` - optional, WebSocket implementation. use this when your environment does not have a built-in native WebSocket (for example, with NodeJS client)
 
 ### Methods
-#### `subscribe(options, handler) => id`
+#### `subscribe(options, handler) => number`: subscribed to data and returns the subscription id
```
mistic: returns the operation id that identifies the subscription

README.md
```diff
@@ -227,90 +259,55 @@ ReactDOM.render(
...
 - => Returns an `off` method to cancel the event subscription.
 
+### `query(options: OperationOptions) => Promise<ExecutionResult>` : Executes GraphQL operation over the WebSocket
+- `options: OperationOptions`:
+    * `query: string` - GraphQL operation as string or parsed GraphQL document node
+    * `variables?: Object` - Object with GraphQL variables
+    * `operationName?: string` - GraphQL operation name
+    * `context?: any` - Execution context for the operation
+    
+### `close() => void` - closes the WebSocket connection manually
```
mistic: when my PR gets merged this should mention that it closes the websocket manually without firing the tryReconnect, even if its set 😄 #137

PROTOCOL.md
```diff
@@ -0,0 +1,74 @@
...
+- Client creates a WebSocket instance using `SubscriptionsClient` object.
+- Client connected immediatly, or stops and wait if using lazy mode (until first operation execution)
+- Client sends `GQL_CONNECTION_INIT` message to the server.
+- Server calls `onConnect` callback with the init arguments, waits for init to finish and returns it's return value with `GQL_CONNECTION_ACK`, or `GQL_CONNECTION_ERROR` in case of `false` or thrown exception from `onConnect` callback.
+- Client gets `GQL_CONNECTION_ACK` and waits for the client's app to create subscriptions.
+- App creates a subscription using `subscribe` or `query` client's API, and the `GQL_START` message sent to the server.
+- Server calls `onOperation` callback, and responds with `GQL_DATA` in case of zero errors, or `GQL_ERROR` if there is a problem with the operation.
+- Server calls `onOperationDone` if the operation is a query or mutation (for subscriptions, this called when unsubscribing)
+- Server sends `GQL_COMPLETE` if the operation is a query or mutation (for subscriptions, this sent when unsubscribing)
+- Client get `GQL_DATA` and handles it.
```
mistic: This here it's a bit confusing, it should me more explicit that the user will receive first the `GQL_DATA` and only after the `GQL_COMPLETE`. It's almost there but I believe those three sentences are confusing the thing a bit:

1- Server calls `onOperation` callback, and responds with `GQL_DATA` in case of zero errors, or `GQL_ERROR` if there is a problem with the operation.
2- Server sends `GQL_COMPLETE` if the operation is a query or mutation (for subscriptions, this sent when unsubscribing)
3-Client get `GQL_DATA` and handles it.

PROTOCOL.md
```diff
@@ -0,0 +1,74 @@
...
+
+For subscriptions:
+- App triggers `PubSub`'s publication method, and the server publishes the event, passing it through the `subscribe` executor to create GraphQL execution result
+- Client receives `GQL_DATA` with the data, and handles it.
+- When client unsubscribe, the server triggers `onOperationDone` and sends `GQL_COMPLETE` message to the client.
+
```
mistic: It might be worth it adding a small section referring the *deprecated* messages of the protocol 😄

PROTOCOL.md
dotansimha: What do you mean? The `GQL_COMPLETE` is send only when the `AsyncIterator` (or `Observable`) will trigger `complete` callback. the `GQL_DATA` will arrive before `GQL_COMPLETE`. Am I wrong?

dotansimha:
@mistic Thanks for the review! Fixed everything

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+  type: string;
+}
+```
+
+### Client -> Server
+
+#### GQL_CONNECTION_INIT
+Client sends this message after plain websocket connection to start the communication with the server
+
+This message also triggers the `onConnect` callback on the server.
```
DxCx: Im not sure this is the right place for such comment,
maybe we need a seperate documentation about events,
but this is Implementation specific and not Protocl specific info.

This ofcourse applies to the rest of the document as well.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `payload: Object` : optional parameters that the client specifies in `connectionParams`
+
+#### GQL_CONNECTION_TERMINATE
```
DxCx: Please move below stop, makes more sense with
CONNECTION INIT => START => STOP => TERMINATE.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+    * `context?: any` : Execution context for the operation
+    
+#### GQL_STOP
+Client sends this message in order to stop a running GraphQL operation execution (for example: unsubscribe)
+- `id: string` : operation id
+    
+### Server -> Client
+
+#### GQL_CONNECTION_ERROR
+The server sends this message if `onConnect` callback returns `false` or throws an exception, and after sending this message, the server disconnects the client.
```
DxCx: Again, this comment is implementation specific, the right phrasing should be something like:
`server may responses with this message to the GQL_CONNECTION_INIT from client, indicates the server rejected the connection`

the implementation specific info should come in another document

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+
+It server also respond with this message in case of a parsing errors of the message (which does not disconnect the client, just ignore the message).
+
+- `payload: Object`: the server side error
+
+#### GQL_CONNECTION_ACK
+The server sends this message if `onConnect` callback returns any other value then `false`.
```
DxCx: same thing as above, indicating server acknowledgment.

Also, we need to add a note that the server will send either CONN_ACK or CONN_ERROR but not both.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+    * `operationName?: string` : GraphQL operation name
+    * `context?: any` : Execution context for the operation
+    
+#### GQL_STOP
+Client sends this message in order to stop a running GraphQL operation execution (for example: unsubscribe)
+- `id: string` : operation id
+    
+### Server -> Client
+
+#### GQL_CONNECTION_ERROR
```
DxCx: Missing info: the server should reject messages sent to it before connection was completed.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+
+It server also respond with this message in case of a parsing errors of the message (which does not disconnect the client, just ignore the message).
+
+- `payload: Object`: the server side error
+
+#### GQL_CONNECTION_ACK
+The server sends this message if `onConnect` callback returns any other value then `false`.
+
+#### GQL_DATA
+The server sends this message to transfter the GraphQL execution result from the server to the client.
```
DxCx: Need a relation that this is Response for GQL_START message,
also, an operation should contain at least 1 DATA per operation, and may contain more then one.

DxCx: Also, this response will be sent only after GQL Engine started execute state.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+- `id: string` : ID of the operation that was successfully set up
+- `payload: Object` : 
+    * `data: any`: Execution result
+    * `errors?: Error[]` : Array of resolvers errors
+
+#### GQL_ERROR
+Server sends this message upon failing to **register** or **execute** a GraphQL operation (resolver errors are part of `GQL_DATA` message, and will be added as `errors` array)
```
DxCx: no need to state register, also, failing to execute is confusing because of the `execute` step of graphql.
this should be any error that happens in the server casued by the operation and before execute step started.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+- `payload: { errors: Array<Object> }` : payload with an array of errors attributed to the subscription failing on the server
```
DxCx: Payload is 1 error and not errors array.
multiple errors can happen only once execute step has started and will be delivered as GQL_DATA errors.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+- `id: string` : operation ID of the operation that failed on the server
+
+#### GQL_COMPLETE
+Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive (for queries and subscription, it called immediatly, and for subscriptions it called when unsubscribing)
```
DxCx: for query and mutation, not subscription.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+- `id: string` : operation ID of the operation that completed
+
+#### GQL_CONNECTION_KEEP_ALIVE
+Server message sent periodically to keep the client connection alive.
+
+The client starts to considerer the keep alive message only upon the first received keep alive message from the server.
+
+### Messages Flow
```
DxCx: Would be great to add here (Maybe in the future and not as part of this PR) a visual representation of the states.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+
+This is a demonstration of client-server communication, in order to get a better understanding of the protocol flow:
```
DxCx: Let's split that into states:
Session init state => GQL_CONNECTION_INIT (ERROR OR ACK)
connected phase => Server accepts GQL_START/GQL_STOP
then foreach Operation recived.
and finally terminating the connection..

Also, there are implementation specific info here.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
+# GraphQL over WebSocket Protocol
```
DxCx: Missing intro section to convert terms between packet names and actual strings on the network.

PROTOCOL.md
DxCx: Missing an actual example on how the protocol packets looks like (JSON Format for example)
that is expected to be in a protocol specification.

PROTOCOL.md
```diff
@@ -0,0 +1,99 @@
...
+- Client get `GQL_DATA` and handles it.
+- Server calls `onOperationDone` if the operation is a query or mutation (for subscriptions, this called when unsubscribing)
+- Server sends `GQL_COMPLETE` if the operation is a query or mutation (for subscriptions, this sent when unsubscribing)
+
+For subscriptions:
+- App triggers `PubSub`'s publication method, and the server publishes the event, passing it through the `subscribe` executor to create GraphQL execution result
+- Client receives `GQL_DATA` with the data, and handles it.
+- When client unsubscribe, the server triggers `onOperationDone` and sends `GQL_COMPLETE` message to the client.
+
+### Deprecated Protocol
```
DxCx: IMO this should be on README.Md as another link and not in here.

PROTOCOL.md
dotansimha: What do you mean? The server does not reject messages if they are sent before the init phase, it just waits for the `initPromise` to resolve.

PROTOCOL.md
dotansimha: Can you explain?

dotansimha:
@DxCx I made some changes. what do you think ?

PROTOCOL.md
DxCx: rough example: :)
Client initates connection
C => `{"type": "connection_init", payload: "..." }`
Server may reject the connection:
S => `{"type": "connection_error", payload: "Permission denid"}`
or accept it with 
S => `{"type": "connection_ack"}`
etc...
like add for each packet type a json example :)

PROTOCOL.md
DxCx: alright then describe this behavior.. if that's the inteded one..
i just mean that what happens in this state should be defined.

PROTOCOL.md
```diff
@@ -18,52 +21,54 @@ export interface OperationMessage {
...
 - `id: string` : operation id
     
+#### GQL_CONNECTION_TERMINATE
+Client sends this message to terminate the connection.    
+    
 ### Server -> Client
 
 #### GQL_CONNECTION_ERROR
-The server sends this message if `onConnect` callback returns `false` or throws an exception, and after sending this message, the server disconnects the client.
+The server may responses with this message to the `GQL_CONNECTION_INIT` from client, indicates the server rejected the connection.
```
DxCx: respond instead of respones

PROTOCOL.md
```diff
@@ -18,52 +21,54 @@ export interface OperationMessage {
...
 
 It server also respond with this message in case of a parsing errors of the message (which does not disconnect the client, just ignore the message).
```
DxCx: let's discuess about this behavior, i think if error occured on init state then we should just close the connection and let the client reset it's own state..

PROTOCOL.md
```diff
@@ -18,52 +21,54 @@ export interface OperationMessage {
...
 
 - `payload: Object`: the server side error
 
 #### GQL_CONNECTION_ACK
-The server sends this message if `onConnect` callback returns any other value then `false`.
+The server may responses with this message to the `GQL_CONNECTION_INIT` from client, indicates the server accepted the connection.
+
+The server will response only with `GQL_CONNECTION_ACK` or `GQL_CONNECTION_ERROR` to `GQL_CONNECTION_INIT` message.
```
DxCx: maybe move this paragraph up to `GQL_CONNECTION_INIT` section

PROTOCOL.md
```diff
@@ -18,52 +21,54 @@ export interface OperationMessage {
...
 #### GQL_ERROR
-Server sends this message upon failing to **register** or **execute** a GraphQL operation (resolver errors are part of `GQL_DATA` message, and will be added as `errors` array)
-- `payload: { errors: Array<Object> }` : payload with an array of errors attributed to the subscription failing on the server
+Server sends this message upon a failing operation, before the GraphQL execution, usually due to GraphQL validation errors (resolver errors are part of `GQL_DATA` message, and will be added as `errors` array)
+- `payload: Error` : payload with the error attributed to the operation failing on the server
 - `id: string` : operation ID of the operation that failed on the server
 
 #### GQL_COMPLETE
-Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive (for queries and subscription, it called immediatly, and for subscriptions it called when unsubscribing)
+Server sends this message to indicate that a GraphQL operation is done, and no more data will arrive (for queries and mutations, it called immediatly, and for subscriptions it called when unsubscribing)
```
DxCx: `for queries and mutation, it called immediately` that's not accurate as this will make the implemantation of reactive graphql invalid.

PROTOCOL.md
```diff
@@ -75,12 +80,23 @@ The client starts to considerer the keep alive message only upon the first recei
 
 This is a demonstration of client-server communication, in order to get a better understanding of the protocol flow:
 
+
 - Server configured with GraphQL schema (and `PubSub` implementation when using subscriptions)
 - Client creates a WebSocket instance using `SubscriptionsClient` object.
+
```
DxCx: Let's separate this part into example.MD or something like that as this is implementation related and not only protocol. this is a practical example of how this server works.

Urigo:
thank you @dotansimha !

PR #135, opened 2017-05-18
fix(NA): fixed infinity reconnects when server accepts connections bu…

mistic:
Guys ( @Urigo @DxCx @helfer @dotansimha @NeoPhi  )  this pull request is a fix for the following situation, otherwise will get an infinity reconnect loop:

1- client connects to the server normally with `reconnect = 3` and `reconnectionAttemps = 2` 
2- for some reason the server closes the connection
3- client tryReconnect as specified and servers accepts the basic connection, however don't send the `GQL_CONNECTION_ACK` and as it's in an internal error state closes the connection again
4- loop of point 3 infinity times. 

So in order to solve this the `backoff.reset()` and the `this.reconnecting = false` only occurs when `GQL_CONNECTION_ACK` is received.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -371,7 +371,7 @@ export class SubscriptionClient {
 
     const delay = this.backoff.duration();
     setTimeout(() => {
-      this.connect(true);
+      this.connect();
     }, delay);
   }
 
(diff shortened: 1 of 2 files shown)


Urigo:
good catch @mistic

PR #137, opened 2017-05-18
feat(NA): force close without try reconnect

mistic:
Guys ( @Urigo @DxCx @helfer @dotansimha @NeoPhi ) this pull request is a fix for the following situation:

1- Client connects to the server with `reconnect=true` to prevent for example temporary network failures.
2- Due to a necessary situation, for example a logout, the user of this package wants to close the client.
3- In this situation, even with `reconnect=true`, the `tryReconnect` method shouldn't be called.


dotansimha:
👍

PR #138, opened 2017-05-19
refact(NA): onsubscribe middleware pr

mistic:
Guys ( @Urigo @DxCx @helfer @dotansimha @NeoPhi ) this PR is to refact somethings that were introduced by #78 

However I also want to raise up the discussion about the middleware thing that @DxCx also comments on #78.

Do we wanna be able to use a middleware and always send the context over the network in every request? Isn't this a severe security issue? Won't we be able to accomplish the exact same use case described on #78 sent the authtoken in the connectionParams with the new `lazy` option enabled?

In my opinion this should't be released until we got a really good conclusion. However I haven't removed that in this PR, I had only refact somethings and also fixing a bug.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -36,10 +36,6 @@ export interface Operations {
   [id: string]: Operation;
 }
 
-export interface PendingOperations {
-  [id: string]: boolean;
-}
-
 export interface Middleware {
   applyMiddleware(options: OperationOptions, next: Function): void;
 }
@@ -78,7 +74,6 @@ export class SubscriptionClient {
   private wasKeepAliveReceived: boolean;
   private checkConnectionTimeoutId: any;
   private middlewares: Middleware[];
-  private pendingOperations: PendingOperations;
 
   constructor(url: string, options?: ClientOptions, webSocketImpl?: any) {
     const {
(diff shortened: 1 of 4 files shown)


NeoPhi:
I agree that we should not be using a client supplied context value blindly for executing the GraphQL query. The ability to specify an `onSubscribe` or `onOperation` which is passed the parsed message would allow for this same extension point.

mistic:
I agree with you @NeoPhi. We'll be able to support the use case using `onOperation` and/or with `connectionParams` + `lazy=true`. If all the others agree I can make one last commit on top of that one removing the middleware code and tests and also update the docs on #134

NeoPhi:
Just to be clear, the client side middleware is useful. I think the only change is that we should not be using `parsedMessage.payload.context` when creating `baseParams`.

srtucker22:
Hey,

Sorry about the security issue -- I've been validating the context on every subscribe and in haste to publish, didn't consider when ppl wouldn't.

Regarding context, I was hoping to keep the connection open while using an alternative context rather than disconnecting/reconnecting if context changes, but it's not a big deal. There could be a way to let context get modified only if it passes through the server's `onOperation` where that function has the authority to update context, but if that still bothers you, we don't need to do it.

I also think the middleware stuff could be useful on the client for other transformations aside from context.

dotansimha:
@mistic 
I agree about the `context` issue, and I think that middlewares are an important addition.
Maybe we can just remove `context`, and enforce creating the context only in server side (like @NeoPhi suggested).

mistic:
@NeoPhi @dotansimha @DxCx what do u think now?

NeoPhi:
LGTM

Urigo:
thank you very very much @mistic !

Issue #161, opened 2017-06-03
Reconnect after closing connection on client


srtucker22:
I'm interested in using the `lazy` option to connect only once the user is authenticated and disconnect (`wsClient.close()`) when they log out, so there are only authenticated connections. However, if someone logs in and subscribes from the same client after `wsClient.close()` has been called, I think it would make sense to be able to reconnect with the same `wsClient`. As [the code stands right now](https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L161 ) I'm not sure this is possible. It [seems like](https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L445) a client was intended to be reusable after `close()`.

Maybe there's a better way to do what I'm trying? Maybe we could just change that conditional to check whether the client is null or closed? Maybe this will break other internal flags and things?

Let me know what's good.

carlosbergfeld:
Also very interested in this! Although I think a new websocket client must be created to reconnect. Maybe `client` could just be set to `null` after closing the connection?

dotansimha:
@carlosbergfeld @srtucker22 
I guess that adding:
```typescript
client = null;
```

here: https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L126

Will solve this issue? or i'm missing something?

srtucker22:
I mean, I think it should be fine, but just wanted to double check with you guys first because there might be flags that need resetting or something. If client is null, for example, `wsClient.status === 0 ` which breaks a couple tests which expect `3`. That was an easy one to spot, but there might be more tweaks idk about

dotansimha:
@srtucker22 
Yeah it makes sense, I think we should also change `status()` to return `CLOSED` (3) when the `client` is `null`.

Please take a look at this PR: https://github.com/apollographql/subscriptions-transport-ws/pull/164 
Do you think this solution solves this issue?

srtucker22:
Wow, so fast! Should solve it. I'll throw up a PR if something else pops up when I'm using it.

Thanks!

Urigo:
released in 0.7.1

PR #164, opened 2017-06-04
Fix for reconnect after manual close

dotansimha:
Closes https://github.com/apollographql/subscriptions-transport-ws/issues/161


Urigo:
great work @dotansimha !

PR #179, opened 2017-06-12
Don't skip opId === 0

cvle:
I noticed that the first subscription is never unsubscribed.
Turns out that the first operation id is zero and that the `if (opId) {`
check is the culprit.

Fixes: #180

TODO:

### DIFF
--- a/src/adapters/subscription-manager.ts
+++ b/src/adapters/subscription-manager.ts
@@ -89,7 +89,7 @@ export const executeFromSubscriptionManager = (subscriptionManager: Subscription
 
         if (subIdPromise) {
           subIdPromise.then((opId: number) => {
-            if (opId) {
+            if (!isNaN(opId) && opId >= 0) {
               subscriptionManager.unsubscribe(opId);
             }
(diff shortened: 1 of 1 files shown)


NeoPhi:
Great catch @cvle. Thank you for the PR!

cvle:
@NeoPhi Awesome! Can you make a release? This was the source for a log of warnings on our side. Would be great to get rid of them :-)

NeoPhi:
@dotansimha or @Urigo See above.

cvle:
@NeoPhi I saw that recently the typings of `operation id` was changed to a `string`. I didn't know. The typings should be adapted and the check should be changed to sth like: `if (opId || opId === 0)`.

cvle:
`https://github.com/apollographql/graphql-subscriptions` still has `number` typings for opIds.

dotansimha:
@cvle where are the `number` for `opIds`?

Urigo:
thanks @cvle ! released in `0.7.3`

PR #196, opened 2017-06-30
chore(lodash): require specific lodash/assign module

WesleydeSouza:
Importing from `lodash` loads the entire lib into memory unnecessarily.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ### vNEXT
+- Require specific lodash/assign module instead of entire package, so memory impact is reduced [PR #196](https://github.com/apollographql/subscriptions-transport-ws/pull/196)
 
 ### 0.7.3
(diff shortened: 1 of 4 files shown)


WesleydeSouza:
It seems the build fail is not related to this PR changes.

mistic:
@WesleydeSouza I think my pr should fix this build fail on your pr :)

WesleydeSouza:
@mistic yeah, for sure the browser build is now massively smaller (~83 kB). 

And it seems the intended package was in fact `lodash.assign`, the previous require will fail if no library depends on `lodash` (which is fine for people that use `webpack` since `async` depends on `lodash` 😛 )

I've switched to that, but it seems I've done something wrong with the type declarations and test is failing again, I'm not very familiar with TypeScript.

dotansimha:
@WesleydeSouza why changing `import` to `require` ? Your typings looks fine and I think you can use `import` here.

dotansimha:
@WesleydeSouza never mind, got it :) 

Thanks @WesleydeSouza !

PR #197, opened 2017-06-30
fix(NA): Non forced close without connection terminate and also some connection's flow improvements

mistic:
@Urigo @dotansimha @DxCx 

This PR solves the following issue: usually when a backend receives a `connection_terminate` will invalidate the previous credentials so we should only send connection_terminate in a `forceClose`.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -73,7 +73,8 @@ export class SubscriptionClient {
   private closedByUser: boolean;
   private wsImpl: any;
   private wasKeepAliveReceived: boolean;
-  private checkConnectionTimeoutId: any;
+  private tryReconnectTimeoutId: any;
+  private checkConnectionIntervalId: any;
   private middlewares: Middleware[];
 
   constructor(url: string, options?: ClientOptions, webSocketImpl?: any) {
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -122,12 +124,37 @@ export class SubscriptionClient {
...
+  public close(isForced = true, closedByUser = true) {
     if (this.client !== null) {
-      this.forceClose = isForced;
-      this.sendMessage(undefined, MessageTypes.GQL_CONNECTION_TERMINATE, null);
-      this.client.close();
+      this.closedByUser = closedByUser;
+
+      if (isForced) {
+        if (this.checkConnectionIntervalId) {
+          clearInterval(this.checkConnectionIntervalId);
```
DxCx: Should set this with undefined or null

src/client.ts
```diff
@@ -424,15 +456,37 @@ export class SubscriptionClient {
   }
 
   private checkConnection() {
-    this.wasKeepAliveReceived ? this.wasKeepAliveReceived = false : this.close(false);
+    this.wasKeepAliveReceived ? this.wasKeepAliveReceived = false : this.close(false, true);
```
DxCx: Please convert into full if and not shortend

PR #208, opened 2017-07-06
fix(NA): install library from branch

### LINKED ISSUE #147: It's not possible to install this module from master branch
Hi guys ( @Urigo @DxCx @helfer @dotansimha @NeoPhi ),

I tried to install this package as a dependency in a personal project from the master branch `apollographql/subscriptions-transport-ws#master` but it completely fails. I've already spotted a lot of things that are blocking this feature. What u think about that?

mistic:
@Urigo @dotansimha @DxCx this PR closes #147 and makes possible to install this library from master branch through npm. I think that is a really needed feature for example to test a feature/bug fix without waiting for the oficial release. Please take a quick look reviewing it.


DxCx:
hey @mistic did you verify that the published package is still installable?
you are doing `rimraf src` on prepublishOnly but postinstall has compile so..

mistic:
No @DxCx but I think you're right. Should be enough to make a prepublish with compile and rimraf :) prepublish only and postinstall could be removed :)

dotansimha:
That's awesome!! Thanks @mistic !

PR #218, opened 2017-07-08
fix(NA): add a return after first post-install when it should install dev dependencies.

mistic:
@Urigo @dotansimha @DxCx this is only a small fix so when installing from a branch it doesn't repeat the compile job for 2 times.


Urigo:
thanks @mistic

Issue #220, opened 2017-07-10
after upgrade 0.7.3 to 0.8.0 error message


dohomi:
Hello,
after I upgraded the package from 0.7.3 => 0.8.0 I receive following error in the console:
```
VM7914:37 WebSocket connection to 'wss://subscriptions.graph.cool/v1/xxx' failed: Connection closed before receiving a handshake response
```
This happens on every connection I try to establish
Any idea how to solve this?

mistic:
Are u using `SubscriptionsServer`? Can u post here your Websocket frames?

dohomi:
```
import 'isomorphic-fetch'
import { ApolloClient, createNetworkInterface } from 'apollo-client'
import { SubscriptionClient, addGraphQLSubscriptions } from 'subscriptions-transport-ws'
import { getTokenFromStorage } from './auth-helpers'
const apolloUri = 'https://api.graph.cool/simple/v1/xx'
const wsUri = 'wss://subscriptions.graph.cool/v1/xx'

export default function () {
  const ssr = process.SERVER_BUILD
  const networkInterface = createNetworkInterface({
    uri: apolloUri,
    transportBatching: true,
    opts: {
      credentials: 'same-origin'
    }
  })

  let wsClient
  let networkConf
  // let initialState
  if (!ssr) {
    wsClient = new SubscriptionClient(wsUri, {
      reconnect: true
    })
  }

  if (wsClient) {
    networkConf = addGraphQLSubscriptions(
      networkInterface,
      wsClient
    )
  } else {
    networkConf = networkInterface
  }

  networkConf.use([{
    applyMiddleware (req, next) {
      if (!req.options.headers) {
        req.options.headers = {} // Create the header object if needed.
      }
      const token = getTokenFromStorage(req)
      req.options.headers.Authorization = `Bearer ${token}`
      next()
    }
  }])

  return new ApolloClient({
    networkInterface: networkConf,
    dataIdFromObject: r => r.id,
    ...(ssr ? {
      ssrMode: true
    } : {
      connectToDevTools: process.env.NODE_ENV !== 'production'
    })
  })
}

```

mistic:
@dohomi Please post here your websocket frames information (the exchange of messages between your connect_init, the received connection_ack and your first keep alive message). Also are u able to connect and use it normally and this is only a warning message, or u are having troubles making the data communications?

dohomi:
[...]

PR #223, opened 2017-07-11
fix(NA): send first keep alive message right after connection ack

mistic:
@Urigo @dotansimha @DxCx this small commits fixs a bug that I found in sequence of helping a user in the issue #220 . We should send the first KA message right after the connection ack in the server side.


dotansimha:
Great job @mistic ! Thank you :)

PR #239, opened 2017-07-27
fix(NA): call operation handler before delete the operation on operation complete

mistic:
@Urigo @DxCx @dotansimha this should fix the complete flow for an operation.


DxCx:
LGTM! 👍

Urigo:
Great job @mistic !

PR #241, opened 2017-07-30
Add missing validation step

### LINKED ISSUE #213: Add SubscriptionServer query resolution function
Currently, SubscriptionServer uses `execute` from `'graphql'` under the hood and in the docs.

If there is a syntax error in a query, it won't be caught because `execute` just executes the query, it doesn't check for syntax error. This is the [issue](https://github.com/apollographql/subscriptions-transport-ws/issues/186) I had when using default implementation.

`graphql` package has a function `graphql` [link](https://github.com/graphql/graphql-js/blob/master/src/graphql.js#L68) that is used to check for [syntax and validation errors](https://github.com/graphql/graphql-js/blob/master/src/graphql.js#L114) and only then runs `execute` function.

Apollo `graphql-server` has also its own [function](https://github.com/apollographql/graphql-server/blob/6d8047f5dd6001918c9c8b91657415fba6058fbc/packages/graphql-server-core/src/runQuery.ts#L62) that also checks for [syntax and validation errors](https://github.com/apollographql/graphql-server/blob/6d8047f5dd6001918c9c8b91657415fba6058fbc/packages/graphql-server-core/src/runQuery.ts#L101) and has additional options like `formatError `.

Current `subscriptions-transport-ws` implementation is incomplete because it doesn't have this checks. So, by default `graphql` method should be used instead of just `execute` or a custom function should be provided with this checks (optionally with options like `formatError`).

DxCx:
I've found out that server missed validation step.
Added it at the same API as Apollo-Server is:
https://github.com/apollographql/apollo-server/blob/master/packages/apollo-server-core/src/graphqlOptions.ts#L25

Closes #213 

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Changelog
 
 ### vNEXT
+- Add Validation step to server [PR #241](https://github.com/apollographql/subscriptions-transport-ws/pull/241)
+- Call operation handler before delete the operation on operation complete [PR #239](https://github.com/apollographql/subscriptions-transport-ws/pull/239)
 
 ### 0.8.1
 - Send first keep alive message right after the ack [PR #223](https://github.com/apollographql/subscriptions-transport-ws/pull/223)
 - Return after first post-install when it should install dev dependencies [PR #218](https://github.com/apollographql/subscriptions-transport-ws/pull/218)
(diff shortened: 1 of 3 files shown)


mistic:
👍 LGTM

Urigo:
thank you so much @DxCx !  good to have you back

PR #261, opened 2017-08-30
Adds Support for Graphql ^0.11.0

### LINKED ISSUE #259: Update to accept promise returned from subscribe with graphql 0.11.x
The subscribe function from graphql returns a promise that resolves to an AsyncIterator.
This is different than was in 0.10.5, and thus breaks with an exception in server.ts line 431 with and exception: Cannot read property 'then' of undefined

Normally, this wouldn't be an issue, but the dependencies allow for 0.11.1 to be installed.
    "graphql": "^0.10.1"

When I cloned and installed this repo, it only installed 0.10.5. But, since I added subscriptions to my library after the fact, graphql didn't get downgraded or a warning to that effect didn't occur,  this became a serious pain.

How to reproduce:
Take any existing example and run with graphql 0.11.x and it will fail.

Also, the tests fail in 0.11.1

DxCx:
This PR **adds** support for GraphQL ^0.11.0
^0.10.0 will still work.

**BREAKING CHANGE**: Remove support for `SubscriptionManager`
Related to:
https://github.com/apollographql/graphql-subscriptions/pull/104

TODO:

Closes #259


edvinerikson:
What's the status of this? No one able to review it? @DxCx

DxCx:
@edvinerikson there is still issue with `graphql-subscriptions` which apollo team needs to decided either remove from dependancies or upgrade as well..
also, as you can see there are older PRs which havn't been reviewed as well.. im waiting as well..

edvinerikson:
Alright, I see. Do you know what the dependency is inside of graphql-subscriptions?

I pinged @stubailo on Slack with the link to my PR and mentioned the low activity. I have not get any reply yet though (20min ago or something).

I really need this change because of the async subscribe function.

DxCx:
only subscription manager interface which i think needs to be just moved inside this package until deprecated or just be deprecated as soon as possible.
but there is no reason of `graphql-subscriptions` to be a dependancy here.

edvinerikson:
I agree. I just migrated away from it when I upgraded. Not sure in which version it got deprecated but making a major bump and removing it shouldn't be too hard(?).

stubailo:
I think @NeoPhi will also take a look soon, let's try to get some next steps before the end of the week.

zackify:
@stubailo anything I can do to help get this out quicker? I was stuck for a long time today until finding this haha

NeoPhi:
I'm in favor of merging this in and having the needed changes to `graphql-subscriptions` made. There has been a lot of confusion about the deprecation warning so adding 0.11 support as part of a version bump on this project and the other one feels like the right next steps.

edvinerikson:
Something I would like to have fixed too is the error handling. Not in this
PR though. I spent quite a few hours trying to debug why it didn't work and
it turns out the library catches its own errors and sends straight to the
client without any way of interception on the sever. E.g. Only thing I had
to go on was "cannot read property .then of undefined" without stack trace.
I didn't even know if it was a server or client error. ¯\_(ツ)_/¯ What's
your thoughts about this?

NeoPhi:
@DxCx Sorry if my comment was unclear. I'm not advocating that `graphql-subscriptions` be kept as a dependency. Instead I'm saying that if we remove the dependency here we should also update `graphql-subscriptions` to remove the deprecated `SubscriptionManager` to avoid any confusion.

DxCx:
Oh, that is absolutely correct. :)

DxCx:
will need a new version of `graphql-subscriptions` so CI won't fail on node 4
but accept for that it looks ready for review to me.

NeoPhi:
@DxCx `graphql-subscriptions` v0.5.0 has been released.

package.json
```diff
@@ -34,16 +32,20 @@
...
+    "@types/graphql": "^0.11.2",
     "@types/lodash": "^4.14.68",
     "@types/mocha": "^2.2.41",
     "@types/node": "^8.0.8",
     "@types/sinon": "^2.3.0",
     "@types/ws": "^3.0.0",
     "chai": "^4.0.2",
-    "graphql": "^0.10.1",
+    "graphql": "^0.11.3",
+    "graphql-subscriptions": "^0.4.4",
```
dotansimha: @DxCx update to latest :)

src/adapters/subscription-manager.ts
```diff
@@ -1,110 +0,0 @@
-import { SubscriptionManager } from 'graphql-subscriptions';
```
dotansimha: 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

dotansimha:
@DxCx looks great. just need to update `graphql-subscriptions` version and fix CI issue. 
Back then we talked about "V1" of the transport, and the goal was to remove the support for the old code. 
If we removed the `SubscriptionManager` support, do you think we need to remove the support for the older protocol as well? @DxCx @Urigo

DxCx:
we don't need to remove it but i think we can remove all deprecated code..
we can do it now or next step will be to clean them up and release `graphql-transport-ws`

package.json
DxCx: yeah i was working at that since there was a typing still depended.. now it's pushed =)

NeoPhi:
As this is a breaking change release, I'd be in favor of removing the deprecated code.
I think removal of the old protocol should be part of the switch to graphql-transport-ws as @mistic has thoughts on that.

DxCx:
so do we want to remove all the API and keep the protocol legacy layer for the time being?

NeoPhi:
That would be my recommendation.

DxCx:
@stubailo @Urigo @mistic @dotansimha @martijnwalraven would be great to get some advice in here..

stubailo:
I agree with both of you! Let's go ahead and do it.

NeoPhi:
@DxCx Do you have a preference on next steps? My thinking is that the changes to remove the deprecated API are simple and mirror what is happening with the removal of SubscriptionManager. We can then do a breaking change release of subscriptions-transport-ws that gets this package working with 0.11.x. Then as part of the switch to the new package name the old protocol can be removed.

DxCx:
Yeah lets merge this and wait with releasing ill open another PR cleaning
all deprecated api keeping the legacy protocol support

mistic:
@DxCx @NeoPhi I agree with you! In my opinion we should release now a breaking change version without the `SubscriptionsManager` and the legacy protocol support. And that should be the last `subscriptions-transport-ws` release. Next we should release a new package called `graphql-transport-ws`. What do u think about this?

stubailo:
Hmm, why not release the breaking change as the new package right away?

mistic:
Both ways are valid. However I think that releasing the breaking changes (of things already marked as deprecated for 4 months) in the `subscriptions-transport-ws` as it very last released version can help people to instant migrate to the new package with 0 effort. What do u think @stubailo @NeoPhi @DxCx @dotansimha @Urigo ?

DxCx:
@NeoPhi 
this package is already supported by `apollo-link-ws` and works with apollo-client 2 🎉

mistic:
@DxCx I was suggesting almost the same as u! The only difference is that u prefer to keep the old legacy protocol support on the server side in the last version of `subscriptions-transport-ws` right?

mistic:
@DxCx wasn't this the same I suggest? :D 

In order to allow everyone understand and no doubts remaining what we are suggesting is:

- Release the `0.9.0` version of `subscriptions-transport-ws` with the cleaned api, the legacy protocol and without the `SubscriptionsManager`
- Rename the repository (so we can keep the stars and all the other history) and all the documentation for the new `graphql-transport-ws`.
- Clean the old protocol support, update again the documentation and release the first 0.0.1 version of the `graphql-transport-ws` deprecating the old npm `subscriptions-transport-ws` package with a message to use the new one instead. 

What do u think @DxCx @NeoPhi @dotansimha ?

DxCx:
Agreed.
I just think initial version can be 0.1.0

mistic:
yeah @DxCx it can be 👍 

As soon as the others write their thoughts I can proceed with the agreed changes 😄

NeoPhi:
@mistic Sounds good. Let me know if you'd like me to handle any of that.

morajabi:
Whatever are you doing, please add some notes or guides to upgrade... I wasted a few hours because there's a lot of conflict between documentation of Apollo and this package. Thank you for the great work here.

PR #279, opened 2017-09-22
docs(KEEP_ALIVE): updated protocol documentation to explain the correct workflow between connection_init, connection_ack and connection_keep_alive messages

mistic:
This PR fixes a missing information on the protocol documentation that should explains the correct message flow between the messages: connection_init, connection_ack and connection_keep_alive messages.

The one suggested by the current documentation is:

`--> connection_init` - 0s
`<-- connection_ack` - 0s
`<-- ka` .                     - 10s

Which is wrong.

The one suggested by this docs update, and the one that should be used is:

`--> connection_init` - 0s
`<-- connection_ack` - 0s
`<-- ka` .                     - 0s
`<-- ka` .                     - 10s

You can found more information about how I found this, in a discussion between myself and some graphcool guys here graphcool/graphcool#309.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -21,7 +21,7 @@ export interface OperationMessage {
 #### GQL_CONNECTION_INIT
 Client sends this message after plain websocket connection to start the communication with the server
 
-The server will response only with `GQL_CONNECTION_ACK` or `GQL_CONNECTION_ERROR` to this message.
+The server will response only with `GQL_CONNECTION_ACK` + `GQL_CONNECTION_KEEP_ALIVE` (if used) or `GQL_CONNECTION_ERROR` to this message.
 
 - `payload: Object` : optional parameters that the client specifies in `connectionParams`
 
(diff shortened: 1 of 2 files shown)


PROTOCOL.md
```diff
@@ -21,7 +21,7 @@ export interface OperationMessage {
 #### GQL_CONNECTION_INIT
 Client sends this message after plain websocket connection to start the communication with the server
 
-The server will response only with `GQL_CONNECTION_ACK` or `GQL_CONNECTION_ERROR` to this message.
+The server will response only with `GQL_CONNECTION_ACK` + `GQL_CONNECTION_KEEP_ALIVE` (if used) or `GQL_CONNECTION_ERROR` to this message.
```
NeoPhi: While very unlikely it would be possible when keep alive is in use for the server to send `GQL_CONNECTION_ACK` and then `GQL_CONNECTION_ERROR` if for some reason the `GQL_CONNECTION_KEEP_ALIVE` failed to send.

DxCx: if KEEP_ALIVE failed to send, ERROR will be failing as well...

PR #294, opened 2017-10-24
add wasDataReceived

MingFaiYau:
In some case ( Graph Cool ) , it will send KA message every 10s , unless send data.
If the server is keeping send data message , it will trigger reconnect and make some message drop during reconnect.

If we can receive message , it means we are connecting . so 'wasKeepAliveReceived' should also be true.
I am going to add a new variable 'wasDataReceived' means receive data, and checking in checkconnect
It prevent reconnect when the connection still avaliable.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -84,6 +84,7 @@ export class SubscriptionClient {
   private lazy: boolean;
   private closedByUser: boolean;
   private wsImpl: any;
+  private wasDataReceived :boolean;
   private wasKeepAliveReceived: boolean;
   private tryReconnectTimeoutId: any;
   private checkConnectionIntervalId: any;
@@ -471,11 +472,12 @@ export class SubscriptionClient {
   }
 
   private checkConnection() {
-    if (this.wasKeepAliveReceived) {
+    if (this.wasKeepAliveReceived || this.wasDataReceived) {
       this.wasKeepAliveReceived = false;
+      this.wasDataReceived = false;
       return;
     }
-
+  
     if (!this.reconnecting) {
       this.close(false, true);
     }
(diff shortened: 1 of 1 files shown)


mistic:
@MingFaiYau thanks for your contribution. I'm not sure if this should be true. The KA message is expecting to be received every at every KA_TIMEOUT cycle. Not being able to not receiving it means we are not connected. 

Please check this PR #279 which describes the last docs update for this thing, which was also discussed with graph cool guys on graphcool/graphcool#309

MingFaiYau:
@mistic 
HI sir,
Here is we discuss in graph cool forum.
https://www.graph.cool/forum/t/invoking-a-subscription-event-frequently-in-a-short-amount-of-time/1298/7

And here is a testing example
demo site: https://minghiu.github.io/apollo-SubscriptionClient-issue/index-apollo.html2
source code: https://github.com/minghiu/apollo-SubscriptionClient-issue3
console log pic:
image

In my case , i am using Apollo ws with React native and take some log..
I found that graph cool won't send first KA until we get a disconnect.
( So i can subcribe all new messages  ) 
After we got a disconnect ,it send first KA and the tigger the "checkConnectionIntervalId"..
For now , server will send KA every 10s...but if we subcribe a new message..the time for send KA will delay..
Like below result, and we drop 
if  timeout is 10s adn assume we receive first KA at 0s
0s   : 1st   KA - 0s
10s : 2nd KA - 10s
        subcribe a new message... - 15s
20s : apollo tigger reconnect...( 3rd  KA - 25s )

Above result i got in my React native app..and some testing.
It make app drop many messages for realtime chat :(

mrbarletta:
I've been following the graphcool ticket and this PR.

To be honest, I agree with Graphcool that ain't necessary to send a KA packet when data has been sent. (from their point of view)

At least it should be a setting so it will play nice with other websocket libraries/servers like graphcool.

minghiu:
guys. my app has not been able to handling re-connection properly due the issue this PR is trying to address. I vote this PR can go in as a setting. It is quite reasonable not to send ka when there is messing coming through to be honest. 

If @mistic insist on the current ka protocal can you help pushing graphcool to address the issue on their end. https://github.com/graphcool/prisma/issues/309

Thanks!

Issue #301, opened 2017-10-31
Suggestions to improve Promise detection


malditogeek:
Hello!
I wanted to ask for some feedback before opening a PR. I identified an issue with the way `subscriptions-transport-ws` deals with `Promises`. In concrete this line right here: https://github.com/apollographql/subscriptions-transport-ws/blob/06ba1e8d6395089c4de92308bd353e4fc598f19d/src/server.ts#L373

The issue I found is that [Google's Trace Agent](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs) wraps requests in a different implementation of `Promise` and `promiseOrIterable instanceof Promise` fails therefore throwing `'Invalid `execute` return type! Only Promise or AsyncIterable are valid values!'`

Do you have any suggestions about how to improve this? Promises are quiet tricky to detect and unfortunately `lodash` doesn't have a helper yet (https://github.com/lodash/lodash/pull/739)

`graphql` itself has its own function that check if the objects has a `then` function in it:

https://github.com/graphql/graphql-js/blob/1017befa7270cc0f074308056cea6115be498a99/src/execution/execute.js#L1256-L1266

Works for you? I can open a PR with that implementation if you feel is solid.

dflor003:
I've been stuck for two days on the same issue! A fix for this would be much appreciated!

Any reason not to just check that the function is thenable? Perhaps instead of an `instanceof` check, perhaps something like this:

```
function isPromiseLike(obj: any) {
  return !!obj &&  typeof obj.then === 'function' && typeof obj.catch === 'function';
}
```

dflor003:
I was able to verify that swapping out the `instanceof` check for the above check, does indeed fix my problem. PR to come soon.

dflor003:
@malditogeek: I pushed up a branch to test these changes. If you want to try them out ahead of the PR going through, just add it:

via `npm`:
```
npm install --save https://github.com/dflor003/subscriptions-transport-ws.git#test-release
```

or via `yarn`:
```
yarn add https://github.com/dflor003/subscriptions-transport-ws.git#test-release
```

seeden:
I think that this is related to https://github.com/nodejs/help/issues/557
I have same problem with PM2

NeoPhi:
Relaxed check for Promise merged in an released.

PR #304, opened 2017-11-19
Bugfix - Loose promise check

dflor003:
**Summary:** 

There was a check on line 373 in `server.ts`:  `promiseOrIterable instanceof Promise`. This check was causing issues when `Promise` is augmented or polyfilled as stated in #301. This PR loosens the checks by only requiring that the instance has a `.then` and `.catch` function.

In addition to that:
- When I first cloned the repo, it wiped the `src` folder. Looks like this is due to the `prepublishOnly` script `rimraf`ing `src`. Seems like the intent was to exclude `src` from the published package, so I moved it to `.npmignore`. I can remove this if this was not the case.
- There were two linter errors about unused variables when I forked off `master`. Doesn't look related to my code, but I fixed them anyways.

TODO:


src/client.ts
```diff
@@ -416,9 +416,8 @@ export class SubscriptionClient {
     switch (this.status) {
       case this.wsImpl.OPEN:
         let serializedMessage: string = JSON.stringify(message);
-        let parsedMessage: any;
```
dflor003: This was failing linting in master.

src/server.ts
```diff
@@ -235,11 +236,10 @@ export class SubscriptionServer {
   }
 
   private onMessage(connectionContext: ConnectionContext) {
-    let onInitResolve: any = null, onInitReject: any = null;
```
dflor003: This was failing linting in master as well.

dflor003:
Anything keeping this PR from getting merged? Is there anything else you guys would like me to do prior to approval?

src/server.ts
```diff
@@ -370,7 +370,7 @@ export class SubscriptionServer {
                   params.variables,
                   params.operationName);
 
-                if (!isAsyncIterable(promiseOrIterable) && promiseOrIterable instanceof Promise) {
+                if (!isAsyncIterable(promiseOrIterable) && isPromise(promiseOrIterable)) {
```
NeoPhi: I'd be more inclined to pull in and use the https://www.npmjs.com/package/is-promise library since it uses the more standard is thenable check.

NeoPhi:
Sorry for the delay in getting to this code review.

src/server.ts
dflor003: Hmm... yeah, that makes sense. Less code to maintain. I'll make the change in a few.

package.json
```diff
@@ -19,7 +20,8 @@
     "ws": "^3.0.0"
   },
   "scripts": {
-    "compile": "tsc",
+    "clean": "rimraf dist",
```
dflor003: Noticed that even when I removed the old `.ts` files, it would still include old test files when I ran `npm test`. Added this `clean` script to run before compile to make sure we aren't accidentally publishing/testing deleted files that may have snuck in.

PR #318, opened 2017-12-14
Fix dead link in setup.md

detrohutt:
Not sure if I updated CHANGELOG.md correctly. Just let me know if I need to change anything.


NeoPhi:
Thank you for the fix.

PR #327, opened 2017-12-27
update to support graphql 0.12 with new sync return method

NeoPhi:
Internal server behavior change to support new GraphQL 0.12 behavior:

> GraphQL can now execute synchronously if all resolvers are synchronous functions. This means execute() may no longer return a Promise if execution completed synchronously.


package.json
```diff
@@ -33,7 +32,7 @@
     "prepublishOnly": "npm run compile && npm run browser-compile"
   },
   "peerDependencies": {
-    "graphql": "^0.10.0 || ^0.11.0"
+    "graphql": "^0.10.0 || ^0.11.0 || ^0.12.0"
```
DxCx: when i was more active we tried to always be comptiable with latest 2 graphql versions
if thats still the case we need to remove 10,
you are already cleaning the quirks we had to do because of version 10=>11 incomptabilty...

src/server.ts
```diff
@@ -354,18 +353,7 @@ export class SubscriptionServer {
...
-                  executionIterable = Promise.resolve(promiseOrIterable as any as AsyncIterator<ExecutionResult>);
-                } else {
-                  // Unexpected return value from execute - log it as error and trigger an error to client side
-                  console.error('Invalid `execute` return type! Only Promise or AsyncIterable are valid values!');
-
-                  this.sendError(connectionContext, opId, {
-                    message: 'GraphQL execute engine is not available',
-                  });
-                }
+                executionIterable = Promise.resolve(promiseOrIterable);
```
DxCx: hmm..
since we now should support:
`Result`
`Promise<Result>`
`Promise<AsyncIterator<Result>>`
we should rename `promiseOrIterable` to `promiseOrResult`
also, we should `Promise.resolve` only if `!isPromise(..)`
and `executionIterable` to `executionPromise`

Also,
the createAsyncIterable in line 366 is probably not needed anymore =)

package.json
NeoPhi: As `graphql-subscriptions` still supports 0.10 to 0.12, I'm inclined to just add the additional version for now.

src/server.ts
NeoPhi: As `Promise.resolve` handles the case when passed in a Promise, I'm inclined to just use that. I've pushed an updated that always makes execution a Promise which I think makes the handling cleaner.

hlehmann:
Any update on this ? Currently the only solution is to add async every where to fix this.

PR #341, opened 2018-01-17
Emit event when an websocket error occurs.

damour:
**Use case:**
Writing simple (load)testing tools. At that moment there is no simple way to determinate connection errors. Also useful to debug network troubles.

**Could this be a plugin?**
No, we can't subscribe to ws error event directly.

**Is there a workaround?**
Save connection state(to check that connection closed by server) and catch `disconnected` event. But we don't have any error details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ ReactDOM.render(
 #### `unsubscribeAll() => void` - unsubscribes from all active subscriptions.
 
 #### `on(eventName, callback, thisContext) => Function`
-- `eventName: string`: the name of the event, available events are: `connecting`, `connected`, `reconnecting`, `reconnected` and `disconnected`
+- `eventName: string`: the name of the event, available events are: `connecting`, `connected`, `reconnecting`, `reconnected`, `disconnected` and `error`
 - `callback: Function`: function to be called when websocket connects and initialized.
 - `thisContext: any`: `this` context to use when calling the callback function.
 - => Returns an `off` method to cancel the event subscription.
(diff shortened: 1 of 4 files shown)


tim-soft:
Will this allow logging the errors with something like [apollo-link-error](https://github.com/apollographql/apollo-link/tree/master/packages/apollo-link-error)?

timsuchanek:
@tim-soft yes I guess so! Can this PR please taken care of? This is pretty important as right now there is no possibility to catch Websocket errors using the `subscriptions-transport-ws` client.

README.md
```diff
@@ -255,7 +255,7 @@ ReactDOM.render(
 #### `unsubscribeAll() => void` - unsubscribes from all active subscriptions.
 
 #### `on(eventName, callback, thisContext) => Function`
-- `eventName: string`: the name of the event, available events are: `connecting`, `connected`, `reconnecting`, `reconnected` and `disconnected`
+- `eventName: string`: the name of the event, available events are: `connecting`, `connected`, `reconnecting`, `reconnected`, `disconnected` and `websocket_error`
```
mxstbr: Let's call this event `error` instead, no need to specify `websocket` just like in the other events.

README.md
```diff
@@ -285,6 +285,11 @@ ReactDOM.render(
 - `thisContext: any`: `this` context to use when calling the callback function.
 - => Returns an `off` method to cancel the event subscription.
 
+#### `onWebSocketError(callback, thisContext) => Function` - shorthand for `.on('websocket_error', ...)`
```
mxstbr: Same here, no need for the `WebSocket`, let's call this `onError`.

README.md
damour: Renamed.

README.md
damour: Done.

PR #347, opened 2018-02-03
add optional payload to ACK

mattkrick:
Sometimes, it's nice to send a payload with an ACK.
A good example might be a JWT, in case the one the client is using is about to expire.
Another example might be the app version, which allows you to upgrade the client without requiring a refresh (this pairs nicely with the built-in reconnect logic).
 
I think you guys were thinking about doing this, too, because I noticed a payload field for ACKs already exists in the tests: https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/test/tests.ts#L282


mattkrick:
all tests pass locally, gonna close & reopen until tests complete under the 2000ms limit...

nuragic:
@mattkrick How do you send a payload from the server on ACK? Is there any way via the public API?

PR #362, opened 2018-02-22
Updated dependency iterall to add support for flowtype v0.66

julienvincent:
Currently this dependency throws an error with flowtype `v0.66`. A fix has been published to iterall `v1.2.1`


NeoPhi:
Thank you for the update!

Issue #167, opened 2017-06-06
0.7.X changes server keep alive semantics in client


NeoPhi:
The SubscriptionClient document currently includes: `timeout?: number : how long the client should wait in ms for a subscription to be started (default 5000 ms)`. However this is not how the parameter is being used. Instead this timeout is being used in the [`GQL_CONNECTION_KEEP_ALIVE` handler](https://github.com/apollographql/subscriptions-transport-ws/blob/master/src/client.ts#L516) to close the client to server connection if a server keep alive message is not received in that time. As implemented it feels like this is more of an idleTimeout. Additionally in previous versions the server sent keep alive were used to keep the underlying network connection open for long lived but infrequent subscriptions. Given our default server configuration this is also a parameter we will have to override in GraphiQL to ensure it doesn't close based on the default 5 second timeout.

mistic:
good catch @NeoPhi!

sorenbs:
At Graphcool we send the keep-alive message at 30 second intervals which in almost all cases is enough to keep the underlying connection open. As it stands the client with default configuration  disconnects 5 seconds after the first keep-alive message arrives.

From the Apollo Slack subscriptions channel:

 > Is there any specific reason for decreasing the default keep-alive interval to 5 seconds? That is much lower than what I am used to from other systems. Either we have to change our server to send ka messages every 2 seconds (to account for network latency), the keep-alive logic in Apollo has to be much more lenient towards missed ka messages or the default interval in the client has to be increased to something more reasonable (say, 30 seconds)

helfer:
@sorenbs 5s sounds too low for me too. But just to check: it's the server that's sending the keepalive message here and the client that disconnects if it doesn't receive it? Shouldn't the client instead be the one that says when it wants the connection to stay alive? Presumably the server shouldn't care if the client wants to disconnect (unless it wastes resources by reconnecting right after that)?

sorenbs:
In our implementation the server is sending the ka message every 10 seconds no matter what the client does. This is simply to keep the connection open. The client is free to disconnect at any time, either by closing the connection or sending the disconnect message.

As I understand the current client implementation, it by default keeps the connection open forever. Then, when it receives the first ka message, it sets a timer for the specified timeout (default 5 sec) and sends the close connection message if a new ka message is not received before that time.

I'm not sure this client behaviour makes the most sense, but given this behaviour, I would like to see the default timeout increased.

dotansimha:
[...]

PR #368, opened 2018-03-07
change default keep-alive setting from 10s to 30s

zixia:
following the graphcool server setting. See: https://github.com/apollographql/subscriptions-transport-ws/issues/167

The problem described at https://www.graph.cool/forum/t/invoking-a-subscription-event-frequently-in-a-short-amount-of-time/1298/6?u=huan is not solved if the timeout is set to 10s.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ### vNEXT
-- change default timeout from 10s to 30s(https://github.com/apollographql/subscriptions-transport-ws/issues/167)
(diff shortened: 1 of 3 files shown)


mxstbr:
According to #167 the timeout was already increased to 30s, is that incorrect? If so we should definitely ship this!

gijo-varghese:
I think there is a bug

```
link.subscriptionClient.maxConnectTimeGenerator.duration = () => link.subscriptionClient.maxConnectTimeGenerator.max
```
Above code fixed the issue

PR #374, opened 2018-03-17
allow using custom WebSocket server implementation

lgandecki:
Hello, this will allow me to write isolated integration tests for subscriptions.
My next step will be to create apollo-link-schema-with-subscriptions (open to a better name! :) ), so people can simply do:

```javascript
new ApolloClient({
    link: new SchemaLinkWithSubscriptions({
      schema,
      context
    }),
    cache: new InMemoryCache()
  });
```
And be able to use apolloClient as if it was connected to a websocket server (so they can do subscriptions as well)

Currently, with my fork of subscriptions-transport-ws in my test I do:

```javascript
const mockServer = require("mock-socket-with-protocol").Server;
const mockWebSocket = require("mock-socket-with-protocol").WebSocket;
const gqClient = () => {
  const port = Math.random(); 
// using incorrect port here - it doesn't matter 
// since the mocked implementation does string matching to connect client/server anyway,
// there are no actual ports being opened, which is the point ;-) 
  SubscriptionServer.create(
    {
      schema,
      execute,
      subscribe
    },
    { host: "ws://localhost", port },
    mockServer
  );
  
  const wsLink = new WebSocketLink({
    uri: `ws://localhost:${port}`,
    webSocketImpl: mockWebSocket
  });

  return new ApolloClient({
    link: wsLink,
    cache: new InMemoryCache()
  });
};
```

and it works like a charm :)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "docs/themes/meteor"]
-	path = docs/themes/meteor
-	url = https://github.com/meteor/hexo-theme-meteor.git
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Changelog
 
 ### vNEXT
-- allow using custom WebSocket server implementation
+- allow using custom WebSocket server implementation [PR #374](https://github.com/apollographql/subscriptions-transport-ws/pull/374)
+- change default timeout from 10s to 30s [PR #368](https://github.com/apollographql/subscriptions-transport-ws/pull/368)
+- pass `request` (`upgradeReq`) to `ConnectionContext` [PR #369](https://github.com/apollographql/subscriptions-transport-ws/pull/369)
+- pass `ConnectionContext` to `onDisconnect()` as second argument [PR #369](https://github.com/apollographql/subscriptions-transport-ws/pull/369)
 
 ### 0.9.6
 - fix shallow cloning on contexts which are classes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ### vNEXT
+- allow using custom WebSocket server implementation
 
 ### 0.9.6
 - fix shallow cloning on contexts which are classes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### vNEXT
 - allow using custom WebSocket server implementation [PR #374](https://github.com/apollographql/subscriptions-transport-ws/pull/374)
+- added `error` event to handle connection errors and debug network troubles [PR #341](https://github.com/apollographql/subscriptions-transport-ws/pull/341).
+
+### v0.9.7
 - change default timeout from 10s to 30s [PR #368](https://github.com/apollographql/subscriptions-transport-ws/pull/368)
 - pass `request` (`upgradeReq`) to `ConnectionContext` [PR #369](https://github.com/apollographql/subscriptions-transport-ws/pull/369)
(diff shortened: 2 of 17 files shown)


lgandecki:
Hey Max! You are right. Let me tweak this tomorrow morning :) 
Happy to hear you are not opposed to the idea - meaning I (most possibly!) won't have to use a fork.

lgandecki:
@mxstbr I added some docs and examples - please let me know if this is a good direction. I can continue on this tomorrow.  :) Thanks!

mxstbr:
The "works with custom WebSocket implementation" test fails on that latest commit, any clues why?

lgandecki:
As per the failure.. looks like it's a rare timing issue, looks like 200 ms between firing the events and the assertion that the subscription data came back wasn't enough. I only updated Changelog so that couldn't affect the test. I can increase the timeout a bit. A nicer way of doing this might be to instead of having an assertion in a timeout to check that both subscriptions finished:
```javascript
 if (result.data) {
            numTriggers += 1;
                assert.property(result.data, 'userFiltered'); (..)
            if(numTriggers === 2) { 
                done()
            }
}
```

in both of them. This way the test would end the moment both subscriptions came back, instead of waiting a particular amount of time to verify that numOfTriggers is equal 2. I reused the pattern that's been established in the previous tests, but I don't think it's ideal to be honest.

Website part - would you rather it not speak about apollo and have a test that's almost a copy paste of the one that I talked about above? 
Or maybe not even have a page about integration testing, but an example of creating a server and connecting a client to it? 

I thought that the most typical use case would be nice to cover :) but I could change that into a blog post, and have more minimal info here.

Thanks!

lgandecki:
I see, thanks for the feedback. I will take a look tomorrow. I get your point. I was torn here, obviously there is a lot of boiler code that's not related to the new functionality and it introduces a lot of noise..  On the other hand this is a piece of code that you can copy/paste and it works (assuming you install all the dependencies, but that's the easy part). Otherwise it might be a bit tricky to understand what all the parts are and how they connect with each other.

lgandecki:
I finally got some time to clean it up a bit. I hope that's better. 

On a different note - I'm also tempted to work on the tests, they are slow, and can't be parallelized. In mocha watch mode they take ~25 seconds to run, which is a rather suboptimal developer experience. :-) 

If it was up to me I'd:
- split client/server to separate files (so they can run in parallel)
- introduce jest for its fantastic watch mode
- use waitForExpect instead of arbitrary setTimeouts that introduce unnecessary long waits in some places. 

Is this something you guys would consider merging?

NeoPhi:
I'd be happy moving to jest as the testing framework.

mxstbr:
Same, I'd love to actually, but please do it in a separate PR!

NeoPhi:
@mxstbr Yes, looks good.

lgandecki:
Ping :-)

lgandecki:
I have people asking me about this, even on apollo slack - is there any reason why this is not merged yet? Is there a different pattern for doing integration testing for subscriptions? 
I personally use my fork, but would definitely prefer to just be able to install and use the main package..

NeoPhi:
@lgandecki Fell off the radar. Merging and releasing now.

danfernand:
Question please, how do you pass in the constant to publish for in
```pubsub.publish(SOMETHING_CHANGED_TOPIC, "passedId");
```

example subsscription server
```
SubscriptionServer.create(
    {
      schema: gql`
        type SomethingLog {
          log: String
        }

        type Subscription {
          onSomethingLog(
            someId: ID!
          ): SomethingLog
        }
      `,
      execute,
      subscribe
    },
    customServer
  );
```

Im trying to trigger a publish via pubsub but I am not piecing it all together

PR #376, opened 2018-03-19
add link to Google PubSub implementation

jonas-arkulpa:
TODO:


NeoPhi:
I think it would be better to rework this section of the documentation to point to the implementations mentioned at https://github.com/apollographql/graphql-subscriptions instead of trying to maintain the list in two places.

jonas-arkulpa:
@NeoPhi implented with 20954cb. Just ping me if the change doesn't fit your ideas.

PR #388, opened 2018-04-02
Fix link to React example

joshkel:
Also some minor grammatical fixes.


abernix:
Slightly tweaked it to make it a relative link (and be a bit more defensive in the future if we choose to move these), but thanks all around! 👏

Issue #351, opened 2018-02-07
Close websocket if there are no active subscriptions


dobesv:
Does this close the websocket if you aren't subscribed to anything?  For us, at least, it seems like if you ever have a subscription then the apollo client keeps the websocket open indefinitely, even if all the subscriptions are stopped.

timsuchanek:
That is a good question, not that I know of.
This would indeed be a very helpful feature! We would like to use it for server-side subscriptions to have the first-line defense for memory leaks.
It could be very interesting both on the Server and Client.

@NeoPhi as the client seems a major use-case and would also solve our problem, what do you think about adding an option to the Client called `inactivityTimeout`?

I don't have a better name right now, but it would exactly do what @dobesv describes:

*After `inactivityTimeout` milliseconds of active subscriptions count = 0, the client disconnects.*
Default value probably infinity (or 0) to not introduce breaking changes.

I would be happy to contribute that!

NeoPhi:
This feels like a good mirror to the `lazy` flag that triggers waiting to establish the connection until a subscription is needed. The `inactivityTimeout` looks good. A default of 0 to disable or time in ms once no active subscriptions are in place to disconnect. @timsuchanek I'd be happy to review a PR for this if you have time.

divyenduz:
@NeoPhi: Once you have a moment please take a look at the PR that implements this. Please let me know if any changes are needed. Thanks.

divyenduz:
@NeoPhi : Thanks for the merge. I think this can also be closed.

PR #390, opened 2018-04-04
Feature: inactivity timeout and tests

divyenduz:
TODO:

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -71,7 +71,6 @@ export interface ClientOptions {
 export class SubscriptionClient {
   public client: any;
   public operations: Operations;
-  public activeSubscriptions: number;
   private url: string;
   private nextOperationId: number;
   private connectionParams: ConnectionParamsOptions;
@@ -124,7 +123,6 @@ export class SubscriptionClient {
     this.reconnectionAttempts = reconnectionAttempts;
     this.lazy = !!lazy;
     this.inactivityTimeout = inactivityTimeout;
-    this.activeSubscriptions = 0;
     this.closedByUser = false;
     this.backoff = new Backoff({ jitter: 0.5 });
     this.eventEmitter = new EventEmitter();
@@ -146,6 +144,7 @@ export class SubscriptionClient {
   }
 
   public close(isForced = true, closedByUser = true) {
+    this.clearInactivityTimeout();
     if (this.client !== null) {
       this.closedByUser = closedByUser;
 
@@ -174,7 +173,6 @@ export class SubscriptionClient {
 
     let opId: string;
 
-    this.activeSubscriptions += 1;
     this.clearInactivityTimeout();
 
     return {
@@ -370,9 +368,9 @@ export class SubscriptionClient {
   }
 
   private setInactivityTimeout() {
-    if (this.inactivityTimeout > 0 && this.activeSubscriptions === 0) {
+    if (this.inactivityTimeout > 0 && Object.keys(this.operations).length === 0) {
       this.inactivityTimeoutId = setTimeout(() => {
-        if (this.activeSubscriptions === 0) {
+        if (Object.keys(this.operations).length === 0) {
           this.close();
         }
       }, this.inactivityTimeout);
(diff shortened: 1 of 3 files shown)


divyenduz:
Fixes https://github.com/apollographql/subscriptions-transport-ws/issues/351

timsuchanek:
Yeah! This is really cool! Looks good to me!
Is this good to merge @NeoPhi ?

src/client.ts
```diff
@@ -167,6 +174,9 @@ export class SubscriptionClient {
 
     let opId: string;
 
+    this.activeSubscriptions += 1;
```
NeoPhi: This should be next to where `this.operations[opId] =` is done to ensure the reference counting it done correctly. Or possibly the check should just be against `Object.keys(this.operations).length` to avoid needing to do reference counting.

src/client.ts
```diff
@@ -167,6 +174,9 @@ export class SubscriptionClient {
...
+    this.clearInactivityTimeout();
```
NeoPhi: If someone is using `lazy = false` it isn't clear to me that this would then get triggered since a request never comes in.

src/client.ts
```diff
@@ -352,6 +362,23 @@ export class SubscriptionClient {
...
   }
 
+  private clearInactivityTimeout() {
+    if (this.inactivityTimeoutId) {
+      clearTimeout(this.inactivityTimeoutId);
+      this.inactivityTimeoutId = null;
+    }
+  }
+
+  private setInactivityTimeout() {
```
NeoPhi: What should the behavior of this function be as it related to `lazy = false` since `unsubscribe` is the only place it gets called.

src/client.ts
```diff
@@ -352,6 +362,23 @@ export class SubscriptionClient {
...
+    if (this.inactivityTimeout > 0 && this.activeSubscriptions === 0) {
+      this.inactivityTimeoutId = setTimeout(() => {
+        if (this.activeSubscriptions === 0) {
+          this.close();
```
NeoPhi: Should add a test to verify that if the connection is closed due to inactivity, a future subscribe will correctly reopen the connection.

src/client.ts
```diff
@@ -352,6 +362,23 @@ export class SubscriptionClient {
     }
   }
 
+  private clearInactivityTimeout() {
```
NeoPhi: Should there be a call to this if the connection is closed for any other reason, otherwise I could see the reconnect logic fighting with the inactivity timeout and possibly tripping the flags for `closedByUser`.

divyenduz:
@NeoPhi : Thanks for the feedback

1. Updated the code to rely on `Object.keys(this.operations).length` instead of manually counting the references.

2. Called `clearInactivityTimeout` on `close` to avoid it messing up with reconnect logic. 

3. Added the test case to check reconnection after connection was closed due to inactivity. 

Please let me know if any further changes are needed.

src/client.ts
divyenduz: I agree, if a connection is closed (say, manually) but inactivity timeout had started. It might close the connection after `reconnection`.

To get around this, I called `clearInactivityTimeout` inside the `close` function.

src/client.ts
divyenduz: Should this function also test for connection to exist? I think this function is idempotent as is and can be called in any state. Not sure on the impact of `lazy` flag on this? Can you please elaborate.

src/client.ts
divyenduz: Done, we are not counting `activeSubscriptions` anymore and relying on `Object.keys(this.operations).length` to test inactivity.

src/client.ts
divyenduz: Thanks, added the relevant test case.

timsuchanek:
@NeoPhi any news on this? Sorry for bothering, but we really need this to be merged as all people using server-side subscriptions with Yoga have connection leaks right now.

NeoPhi:
@timsuchanek I would contend that if a well behaved client is causing memory leaks in Yoga, that sounds like a server side issue. Long lived idle WebSockets are fine.

PR #431, opened 2018-06-19
fix: use lodash/isString and lodash/isObject instead of lodash.isobject, lodash.isstring

### LINKED ISSUE #430: using lodash instead of lodash.assign, etc. will actually enable smaller webpack builds
the lodash modular build packages like `lodash.assign` etc. are actually larger than you would expect, unable to share code with each other, and waste space in a webpack bundle.

It will save more space in webpack bundles to just
```
import {assign} from 'lodash'
```
and use `babel-plugin-lodash`/`lodash-webpack-plugin` to convert this to
```
import _assign from 'lodash/assign'
```

jedwards1211:
fix #431

lodash modular builds like `lodash.isobject` and `lodash.isstring` can actually be quite large.
By requiring the modules directly from `lodash/isString` etc, we can enable smaller webpack bundles.

TODO:


package.json
```diff
@@ -44,11 +41,12 @@
     "@types/node": "^8.0.8",
     "@types/sinon": "^2.3.0",
     "@types/ws": "^5.1.2",
+    "babel-plugin-lodash": "^3.3.4",
```
NeoPhi: I don't see where this plugin in being enabled?

jedwards1211:
Oh good catch!  I was used to using it in my own projects but then I remembered this is typescript :)

NeoPhi:
Unfortunately this change **increases** the browser/client.js size by 2KB. The currently published version is 68066 while the version produced by this change is 70968, so unless I'm missing something this isn't a net gain.

jedwards1211:
@neophi with what we pack/Babel configuration?

Plus keep in mind that when other parts of the client code are using these same lodash modules, there's greater savings.

jedwards1211:
@neophi I mean, I see you are using we pack for this unpkg build.  But apps that build their own code using we pack should not be using a prebuilt bundle - they should be using the relevant typescript-compiled modules from this package directly.

Btw, John Dalton doesn't plan to release lodash.isobject style packages when he makes a 5.0 release.  He says it is a sort of soft deprecation.

NeoPhi:
Closed in favor of: https://github.com/apollographql/subscriptions-transport-ws/pull/437

PR #437, opened 2018-06-23
fix: use plain JS code instead of lodash.isobject/lodash.isstring

jedwards1211:
TODO:

The use was pretty unnecessary.  People seem to think they need these functions without deeply understanding what they do, and without realizing their primary purpose is to be passed as predicates to functions like `_.some` etc. instead of merely being used in place of `typeof thing === 'string'`.

Note that `isObject` accepts functions, but the code I replaced it with (`foo && typeof foo === 'object'`) does not.  However I think it was accidental to accept a function for `variables` and it would probably be equally accidental if someone is passing a function there.  This code will still accept `RegExp`s, `Arrays`, and non-plain JS objects for `variables` though, just like the current code does :wink:
`isString` accepts not only primitive strings but also the funky `String` wrapper objects, but I don't think anyone uses those intentionally.  And if they do it's probably for some even more funky hack that relies on tacking properties onto the `String` wrapper.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4,8 +4,6 @@ const NativeWebSocket = _global.WebSocket || _global.MozWebSocket;
 
 import * as Backoff from 'backo2';
 import { default as EventEmitterType, EventEmitter, ListenerFn } from 'eventemitter3';
-import isString = require('lodash.isstring');
-import isObject = require('lodash.isobject');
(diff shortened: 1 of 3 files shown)


NeoPhi:
To allow easily switching back I left the method names as is but provided implementations under utils. Thank for you for the research and suggestion. In place versions reduced the client bundle size by 3KB.

PR #428, opened 2018-06-18
Fix websocket reconnection

jpgarcia:
TODO:

- [x] Make sure all of the significant new logic is covered by tests
- [x] Rebase your changes on master so that they can be merged easily
- [x] Make sure all tests and linter rules pass
- [x] Update CHANGELOG.md with your change

PR #439, opened 2018-06-29
Fix duplicate subscription after reconnect

damour:
Same patch as in https://github.com/apollographql/subscriptions-transport-ws/pull/428 but with test.


NeoPhi:
Thank you for adding the test.

PR #348, opened 2018-02-04
Allow connectionParams to be a promise

Rusfighter:
Accept promise as connectionParams function

PR #443, opened 2018-07-13
Allow connectionParams as a Promise

rapzo:
Trying to solve pending #348 issues in order to allow to have `connectionParams` as a Promise.


NeoPhi:
Thank you. I should have a chance to look at this in the next couple of days.

PR #447, opened 2018-07-17
Allow dynamic schema resolution 

nirinchev:
Sometimes, the GraphQL schema needs to be dynamically resolved (e.g. when exposing a database interface). This change allows the schema to be provided either as a static value or as a function of the connection context:

```ts
const subscriptionServer = new SubscriptionServer(
{
    schema: undefined,
    onOperation: async (message, params, socket) => {
        if (await something()) {
            params.schema = schema1;
        } else {
            params.schema = schema2;
        }

        return params;
    },
    ...
}
```

There are no tests for it since it's a fairly trivial change and I wasn't sure if it would be accepted, but I can write some smoke tests if required.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ### vNEXT
+- Allow dynamically specifying/overriding the schema in the object returned from `onOperation` [PR #447](https://github.com/apollographql/subscriptions-transport-ws/pull/447)
 
 ### v0.9.12
 - use lightweight lodash alternatives [Issue #430](https://github.com/apollographql/subscriptions-transport-ws/issues/430)
--- a/README.md
+++ b/README.md
@@ -303,10 +303,10 @@ ReactDOM.render(
 ### `Constructor(options, socketOptions | socketServer)`
 - `options: {ServerOptions}`
   * `rootValue?: any` : Root value to use when executing GraphQL root operations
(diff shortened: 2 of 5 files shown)


nirinchev:
CI seems to be failing with a test that passes locally - I imagine it's possible that this change triggers an underlying race condition with that test. I'll investigate.

NeoPhi:
I think an approach along the lines of apollo-server, would be for the `onOperation` result to optionally return the schema that should be used as it already handles returning the `context`, etc.

nirinchev:
@NeoPhi thanks for the feedback - can you review if the current implementation is aligned with what you had in mind?

NeoPhi:
This is looking great. We will also want updates for the README and docs.

nirinchev:
I updated the readme and the docs - they don't seem to be too verbose, so wasn't sure into how much detail I should go.

NeoPhi:
Thank you very much!

nirinchev:
You're welcome! Do you guys do releases based on fixed time periods or when a certain number of features/bug fixes accumulate? Just wondering what timeframe to give our customers about this making it into an official release 🙂

NeoPhi:
I've been trying to do continuous releases. As such https://github.com/apollographql/subscriptions-transport-ws/releases/tag/v0.9.14 just published.

PR #473, opened 2018-10-09
PROTOCOL.md doc improvements

darrenclark:
I have been looking at integrating the Apollo WebSocket protocol in a non-Node.js project and have been playing around with the GitHunt-API example project.  Figured I'd update `PROTOCOL.md` with some of my findings.

I have a few parts I'm not 100% sure about:

1. In the case of resolver errors, previously the docs said that `GQL_START` might 

    > "also return `GQL_ERROR` with `errors` array, in case of resolvers errors"

    From my testing with the GitHunt sample project, I believe it is actually `GQL_DATA` that is returned with an `errors` array.  Is `GQL_DATA` correct?

2. If a `GQL_START` results in a `GQL_ERROR`, should the client also receive a `GQL_COMPLETE`?  

3. Digging through the code, I see that `GQL_ERROR` (and `GQL_CONNECTION_ERROR`?) usually have a `payload` with a `message` field.  Do you think it's worth me adding it into the docs somehow?

I was wondering if anyone had any insight into the above questions.  I will update the docs to reflect the answers.  Thanks!

**TODO:**

- ~Make sure all of the significant new logic is covered by tests~ (no code changes)
- ~Make sure all tests and linter rules pass~ (no code changes)
- ~Update CHANGELOG.md with your change~ (no code changes)

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -23,10 +23,12 @@ Client sends this message after plain websocket connection to start the communic
 
 The server will response only with `GQL_CONNECTION_ACK` + `GQL_CONNECTION_KEEP_ALIVE` (if used) or `GQL_CONNECTION_ERROR` to this message.
 
+- `type: "connection_init"`
 - `payload: Object` : optional parameters that the client specifies in `connectionParams`
 
 #### GQL_START
 Client sends this message to execute GraphQL operation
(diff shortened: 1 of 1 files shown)


darrenclark:
@meteor-bot done!

PR #492, opened 2018-11-17
Remove labels from GH templates since we're disabling Apollo Bot

hwillson:
Removing all mentions of labelling from this repos custom issue/PR templates (since we're no longer using Apollo Bot).


mxstbr:
Out of curiosity, why did you stop using Apollo bot?

hwillson:
@mxstbr Due to mislabelling. A large percentage of new issues/PRs were being created with incorrect labels, which was adding to the triage overhead across repos.

PR #508, opened 2019-01-30
Add module build

### LINKED ISSUE #500: Distribute ES Modules
When rolling up subscriptions-transport-ws:

```
Error: 'SubscriptionClient' is not exported by node_modules/subscriptions-transport-ws/dist/index.js
```

There's no reason not to distribute browser-friendly es modules.  Typescript can do this easily with `--module "es6"`

Any tool that requires commonjs can *very* easily just use the `esm` loader for pain-free "it-just-works" 'interop'.

bennypowers:
Fixes #500 
TODO:

- [x] Make sure all of the significant new logic is covered by tests (n/a)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules
 .DS_Store
 dist
+module
 coverage
 typings
 npm-debug.log
--- a/package.json
+++ b/package.json
@@ -3,7 +3,9 @@
   "version": "0.9.15",
   "description": "A websocket transport for GraphQL subscriptions",
   "main": "dist/index.js",
-  "browser": "dist/client.js",
+  "module": "module/index.js",
+  "jsnext:main": "module/index.js",
+  "browser": "browser/client.js",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/apollostack/subscriptions-transport-ws.git"
@@ -17,7 +19,7 @@
   },
   "scripts": {
     "clean": "rimraf browser dist coverage",
-    "compile": "tsc",
+    "compile": "tsc && tsc -p tsconfig.module.json",
     "pretest": "npm run compile",
     "test": "npm run testonly --",
     "posttest": "npm run lint",
--- a/tsconfig.module.json
+++ b/tsconfig.module.json
@@ -1,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "es2015",
+    "module": "es2015",
+    "moduleResolution": "node",
+    "sourceMap": true,
+    "noImplicitAny": true,
+    "rootDir": "./src",
+    "outDir": "./module",
+    "allowSyntheticDefaultImports": true,
+    "removeComments": true,
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+    "pretty": true,
+    "declaration": true,
+    "skipLibCheck": true,
+    "lib": ["es6", "esnext.asynciterable"],
+    "types": [
+      "node"
+    ]
+  },
+  "exclude": [
+    "node_modules",
+    "dist"
+  ]
+}


jaydenseric:
For native ESM to work in Node.js `--experimental-modules --es-module-specifier-resolution=node` mode, the package `main` entry needs to be extensionless; in this case it should be changed to `dist`. 

Also, the ESM files need a `.mjs` extension, and they need to be output sibling to the `.js`/CJS files.

morficus:
@jaydenseric both the `main` and `module` entry can live side by side in the Node + npm word.

By default Node will load w/e file is specified by the `main` attribute when looking for CJS modules and will default to the `module` attribute when looking for ESM module. But this is all in experimental mode (with no core group nor community consensus) ([this StackOverflow answer](https://stackoverflow.com/a/42817320/1021300) has some good links about the proposals)

Seeing as how ESM is not a "first class citizen" in Node yet, I would not recommend updating the `main` attribute to point to an ESM. Sounds like a great way to break things for a lot of people. Instead it would be best to make this package available as both a CJS and ESM - which is exactly what this PR would do. 

I for one would love for this to get merged and have this package export all 3 types of modules - CJS (`main`), ESM (`module`) and browser-safe (`browser`)

morficus:
@trevorblades really sorry for the ping, but could you take a look at this please?

jaydenseric:
@morficus

> By default Node will load w/e file is specified by the `main` attribute when looking for CJS modules and will default to the `module` attribute when looking for ESM module.

The `package.json` `module` field which is used by bundlers such as Webpack and Rollup (`jsnext:main` is deprecated) [will never be utilized by Node.js](https://github.com/nodejs/modules/blob/master/doc/announcement.md#es-module-code-in-packages).

The official plan for native ESM in Node.js is available [here](https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md); the final implementation will finally [ship in the next month with Node.js v12.x LTS](https://github.com/nodejs/modules/issues/379).

With all the `--experimental-modules` changes the whole situation for dual CJS/ESM packages has become a mess, in the end [it might not be possible](https://github.com/nodejs/modules/blob/master/doc/plan-for-new-modules-implementation.md#in-progress). Eventually for my own packages I might just ship ESM and drop support for Node.js versions older than v12.x LTS.

In the meantime, [my earlier suggestions](https://github.com/apollographql/subscriptions-transport-ws/pull/508#issuecomment-498925887) are still viable. There are 2 options:

1. Do things the legacy `--experimental-modules` way, which still works both for Webpack, and a specific range of Node.js versions. I'll keeping using this approach in my packages for now to avoid breaking changes.
2. Publishing ESM with the `.js` extension at a `package.json` `module` entrypoint just for bundlers like Webpack so at least you get tree shaking for named exports. This is probably easier and safer.

morficus:
+1 for option 2

aholland-carecloud:
Hi, just curious if anything is blocking this pr as it would allow this package to support all three types of modules and allow for treeshaking for bundlers like webpack? Is this a good solution even if temporarily until node officially supports ESM natively?

trevorblades:
@morficus unfortunately, I'm not the right person to review this PR. I only worked on the docs portion of this repo and am not familiar with the code. There has been some discussion about repo maintenance duties in this thread, if you'd like to get caught up with the state of things: https://github.com/apollographql/subscriptions-transport-ws/issues/777

wighawag:
Hey, I tried that PR. I had some compilation but managed to get it working.
Then when I tried to import it in a `vite` (https://github.com/vitejs/vite) project it failed with

```
[vite] Dep optimization failed with error:
[vite]: Rollup failed to resolve import "utf-8-validate" from "..\node_modules\.pnpm\ws@5.2.2\node_modules\ws\lib\validation.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`rollupInputOptions.external`
```

So I guess the module itself depends on non browser lib ? Any idea?

PR #513, opened 2019-02-24
Inline sources contents in source map files

alexkirsz:
Changes proposed in this pull request:

- Inline sources contents in source map files. 

The current behavior is for `.map` files to refer to the source file. The issue is that the `src` directory is not part of the NPM package, which makes this package's source maps useless as they point to non-existent files. This causes a lot of warnings with `source-map-loader`.

This proposed change makes it so the content of source files gets inlined in the `.map` files, so there is no actual need to read the source files.

Another way to do this would be to bundle the `src` directory within the NPM package.


MrTibbles:
+1 to this getting merged.

MrTibbles:
This is a better solution than #560 IMO

PR #560, opened 2019-05-07
Add src files for source-maps

DanielMSchmidt:
An alternative to this would be #513. Both fix the problem of source maps pointing into the void

This lets the current source maps point to a valid file instead of the void

TODO:

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -8,5 +8,4 @@ dist/test/
 docs/
 npm-debug.log.*
 package-lock.json
-src/
 yarn.lock


niekert:
Would be really nice if this could get merged or some alternative strategies are provided 🙏

meuter:
Any idea when this will be merged? 😅 🙏 (tx!)

MrTibbles:
+1 to this getting merged.

hwillson:
Thanks for this @DanielMSchmidt - we're going to go with https://github.com/apollographql/subscriptions-transport-ws/pull/513 for now.

PR #599, opened 2019-07-11
Hold subscription starts until connection_ack. Fixes #339.

### LINKED ISSUE #339: Client does not wait for connection_ack on reconnection
When the client reconnects after a disconnect, it appears to flush the unsent message queue before receiving 'connection_ack' from the server. This means that there is no guarantee that your server-side onConnect logic will have completed before the first operation is received. This causes an issue if your onConnect method sets the context, as that context will not be present on the operations that are received.

brandonbloom:


brandonbloom:
Actually, as is, this is only a partial fix. Let me dig a bit more.

brandonbloom:
Added a commit containing a test that demonstrates the problem and then passes with the partial fix. Working on a more robust test/fix.

brandonbloom:
Added a failing test: "waits for acks for messages queued between reconnects"

A little more surgery will be necessary to get this test to  pass. Will hopefully have that working sometime next week.

julienvincent:
Awesome, glad a fix is on its way! Just wanted to point out that I am getting this problem on normal connects (not just reconnects) and want to confirm that this covers for that case too?

Screenshot 2019-07-13 at 22 12 46

brandonbloom:
> glad a fix is on its way

Just a heads up: I'm a new contributor, so I can't promise if/when this will be accepted.

> I am getting this problem on normal connects (not just reconnects) and want to confirm that this covers for that case too

Should be.

julienvincent:
> I'm a new contributor

Noted, if this does end up getting blocked I'll be happy to jump in and help out when I get some time myself.

brandonbloom:
OK, I had to do some non-trivial surgery, but I think the result is an improvement. I also made some improvements to the test suite.

@julienvincent I'd encourage you to give this code a try and see if it satisfies your problems. Let me know if you do.

I'm going to test this a bit on our end, but it might take a few days before I can confirm that this fix is stable. Thanks!

julienvincent:
Nice! I gave it a go and seems to be working as advertised.  My testing was pretty much just reload spamming the browser but this issue was happening pretty consistently for me and I can no longer reproduce using your branch.

Thanks for this, hopefully it doesn't take too long to get merged!

brandonbloom:
One potential issue with this change: There is no timeout when waiting for an ack. If an ack never comes, retries won't either. I'll try to add a test and fix for this soon.

arendjr:
Which backend are you guys using? I too am experiencing problems similar to #339 that made me hopeful this PR would solve my woes. I do have a mock backend that I can use for experimenting, and I have built in a delay so that `onConnect()` takes a while to complete before returning and setting the context, and I made sure I explicitly check for the initialized context in my subscription handlers. Still, I cannot reproduce the problem this PR is supposed to solve.

I can see in the Chrome inspector the subscriptions are indeed sent before the `ack` is received, it just appears it doesn't lead to any issues for me because I think the messages are queued probably on the server-side.

arendjr:
To add to this, I have now been able to recreate a scenario where I could break my subscriptions after a reconnect: By making sure my backend rejects the connection in `onConnect()` 50% of the time. Because the subscriptions were already submitted before the `ack` was received, I wouldn't just get a `connection_error` with a "Prohibited connection" message, I would also receive errors for all the subscriptions that had failed because of it. This then canceled the subscriptions on the client, making sure they would be broken, even when a later reconnect was successful.

To avoid these issues, I believe waiting for the `ack` could indeed help. Though I do think that a better retry mechanism on the client-side for the subscriptions themselves could also improve matters. I'll try that next now :)

arendjr:
Okay, so yes: With client-side error handling of subscription errors and retrying them, I can get it to work more reliably. But there's one thing I overlooked: Whenever I receive `connection_error` when creating a connection, all subscriptions are killed and need to be re-established, regardless of whether I received errors on the individual subscription or not.

brandonbloom:
@arendjr I'm using a custom backend. It's true that for a typical websocket server setup, messages will be queued on the backend. However, there are two confounding factors: 1) the client-side queuing mechanism that this PR replaces throws most guarantees of order out the window and 2) relying on the network as a queue doesn't help if you are, like me, terminating web sockets with a gateway and translating messages in to non-blocking requests (like an HTTP POST or a Serverless invocation).

loudmouth:
Any update on this fix? sending `'start'` before sending `'connection_init'` and receiving `'ack'` is an issue i've been grappling with as well and would love to see a fix merged for :-)

devura-se:
@brandonbloom Is this fix fully implemented in your opinion? Or are there more things to be done before it can be merged?

I'm using [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) on the client side with server running on AWS Lambda with [aws-lambda-graphql](https://github.com/michalkvasnicak/aws-lambda-graphql). Client needs to wait for the ack before sending queries, mutations or initializing subscriptions to allow the authentication process to complete before handling the operations.

Haven't yet tried the fix in this PR, but will do.

Shahor:
Hello, what's the progress status on this? 

We can see many people pushing for this to be pushed as they encounter this issue and according to [this comment](https://github.com/apollographql/subscriptions-transport-ws/pull/767#discussion_r514908138) this seems to be the selected solution.

Is there any way I can help moving this forward? We're hitting a wall because of this (I can describe the sequence of things leading to this if you want) and need to move forward and would rather this be fixed upstream.

Please let me know, I'm happy to help.

brandonbloom:
As far as I know, this branch works fine for the folks who I wrote it for, but I'm no longer involved. At this point, I'm not able to contribute more time here, so some core maintainer will have to carry the baton across the finish line.

Shahor:
It seems this is abandonned in favor of https://github.com/enisdenjo/graphql-ws anyway, so I guess you won't have to deal with it anymore.

Thank you anyways for the reply 🙇

PR #615, opened 2019-08-05
Clear WebSocket event listeners on close.

### LINKED ISSUE #581: WebSocket event listeners not cleaned up after WebSocket is closed
After restarting the backend, our client ends up in a continuous reconnection loop.

Our backend is using `gqlgen` for its GraphQL API. The current version implements keep alive, but only sends the "KA" messages on a regular basis after the connection has been set up, default is every 25 seconds. So there is no "KA" directly after the "ACK" only after 25 seconds.

The `SubscriptionClient` in our client is configured to with `reconnect: true` and to use an extended `ws` implementation. `ws` is only extended so we can pass an `Agent` with certificates to the constructor of `ws`, there are no other changes to `ws` logic.

The flow causing our issue as far as I have analyzed it is close to the following:

1. The client, i.e. `SubscriptionClient`, connects to the backend, that is running on the same computer, by creating a new WebSocket, `ws`, and adds an event listener to the WebSockets `onclose` event (as well as other events)
2. Client receives "ACK"
3. After 25 seconds the client receives the first "KA" and starts the `CheckConnectionInterval` that is triggered every 30 seconds from now on
4. A about 10 seconds after the first "KA" I restart the backend
5. The WebSocket detects this and emits a `close` event that triggers the `onclose` event handler
6. The client calls close on the WebSocket and sets it reference to null, but it does not clear the event listeners
7. The client reconnects to the backend, creates and sets up a new WebSocket. Let's call this one "B"
8. Client receives "ACK"
9. Before the "KA" message is received the client triggers the check connection logic, that detects that it has not received a "KA" message and thus closes WebSocket "B" as in 6.
10. WebSocket "B" initiates the closing handshake and sets a timeout that destoys the socket after 30 seconds if the closing handshake fails, i.e. does not get a close event from the socket. And this handshake fails, but I have not figured out why or how. The `onclose` is not triggered yet.
11. The client reconnects to the backend, creates and sets up yet another new WebSocket. Let's call it "C".
12. Client receives "ACK"
13. For the sake of simplicity here let's assume "KA" is received so the check connection interval does not close the connection.
14. However the WebSocket "B"'s closing handshake timeout now triggers and destroys it's socket
15. Since the client has not cleaned up the event listeners from WebSocket "B", WebSocket "B" will now trigger the `onclose` and thus the client will now close WebSocket "C"

This combination of late "KA" messages, the failure of the WebSocket's closing handshake and the event listeners not being properly cleaned up causes our client to end up in a continuous reconnection loop.

client deps:

- [subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws) version 0.9.16
- [ws](https://github.com/websockets/ws) version 7.0.0
- [apollo](https://github.com/apollographql/apollo-tooling) version 2.11.1
- node version 12.4.0

backend deps:

- [gqlgen](https://github.com/99designs/gqlgen) version 0.9.0
- [go] (https://golang.org/) version 1.12.6

tretne:
Resolves #581 

TODO:

- [ ] Rebase your changes on master so that they can be merged easily

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -167,6 +167,10 @@ export class SubscriptionClient {
       }
 
       this.client.close();
+      this.client.onopen = null;
+      this.client.onclose = null;
+      this.client.onerror = null;
+      this.client.onmessage = null;
       this.client = null;
       this.eventEmitter.emit('disconnected');
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Do not send GQL_STOP when unsubscribing after GQL_COMPLETE is received.  <br/>
   [@onhate](https://github.com/onhate) in [#775](https://github.com/apollographql/subscriptions-transport-ws/pull/775)
+- Clear WebSocket event listeners on close.  <br/>
+  [@tretne](https://github.com/tretne) in [#615](https://github.com/apollographql/subscriptions-transport-ws/pull/615)
 
 ### New Features
 


src/client.ts
```diff
@@ -167,6 +167,10 @@ export class SubscriptionClient {
       }
 
       this.client.close();
+      this.client.onopen = null;
```
taras-slapdash, line 170: Hi,

After this change we started getting the following error in our app:

```ts
[TypeError: Cannot set property 'onopen' of null]

e.close (webpack://[name]/node_modules/subscriptions-transport-ws/dist/client.js:126:33)
close (webpack://[name]/node_modules/subscriptions-transport-ws/dist/client.js:477:22)
```

I believe it is due to some race conditioning where two simultaneous calls are made to `.close()`.

Unfortunately, I don't have a repro for this bug so it's tricky to tell exactly what went wrong. 

The call stack starts in the `.onclose` handler (see https://github.com/apollographql/subscriptions-transport-ws/blob/v0.9.18/src/client.ts#L580-L584). 

Any ideas?

dko-slapdash: @tretne I think this `Cannot set property 'onopen' of null` error happens when close() is called synchronously from within close() in a nested way (my guess - during this.unsubscribeAll() call - the subscription handlers may have some cleanups within themselves which close the the client).

The solution would be to either:
1. have one more check before using this.client on L169 and below, or
2. assign null to this.client early (if it's applicable).

ibash: Think @dko-slapdash is right -- I'm able to reproduce this pretty easily by loading my app with the network down.
e.g.
1. Host backend locally
2. Point ngrok at the backend
3. Point the client at ngrok

Refresh the app.

A solution is:

```
const client = this.client
this.client = null
// rest of cleanup ...
```

ibash: https://github.com/apollographql/subscriptions-transport-ws/pull/829

PR #675, opened 2019-10-17
feat(client): add minTimeout option

### LINKED ISSUE #295: Connection timeout strategy causing disconnections
@mistic 

Since the PR that adds the following:
https://github.com/apollographql/subscriptions-transport-ws/blob/v0.8.3/src/client.ts#L407-L416

The default connection timeout starts with 1000ms which isn't enough for low performance servers (for example: free plan in heroku or free plan in `now`).
This causing the client to disconnect if it took more than 1000ms to perform the connection.

For example, with my remote server, I expect the client to connect in about 1500ms or more - in this case I'm seeing one or more disconnections every time I recreate my socket. 

The 1000ms should be configurable, and I think it should be more than 1000ms by default.

apollo-client: 1.9.3
transport: 0.8.3

jedwards1211:
actually fix #295 completely

Does anyone care if I add tests for this?  It's not really "significant new logic"

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 # Changelog
 
-### v0.9.18 (not yet release)
+### v0.9.18 (not yet released)
 
 - Do not send GQL_STOP when unsubscribing after GQL_COMPLETE is received.  <br/>
   [@onhate](https://github.com/onhate) in [#775](https://github.com/apollographql/subscriptions-transport-ws/pull/775)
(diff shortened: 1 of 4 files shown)


jedwards1211:
Please merge soon

jedwards1211:
can I please get an update about when this will get merged?  Am I going to have to resort to monkeypatching (if it's even possible) to fix the spurious #295 disconnect messages we're getting in production?

perrosnk:
When will this be merged?

PR #736, opened 2020-02-08
avoid unhandled rejections when onConnect throws

### LINKED ISSUE #222: UnhandledPromiseRejectionWarning if onConnect rejects or throws
If `onConnect` returns a promise that rejects, or if it throws internally, there is no catch on either `MessageTypes.GQL_START` or `MessageTypes.GQL_STOP` promise handling.

This is throwing a `UnhandledPromiseRejectionWarning`, which will terminate the Node.js process.

```
(node:63913) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Missing jwt
(node:63913) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

OmgImAlexis:
TODO:

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -244,16 +244,35 @@ export class SubscriptionServer {
       const opId = parsedMessage.id;
       switch (parsedMessage.type) {
         case MessageTypes.GQL_CONNECTION_INIT:
+          const errorHandler = (error: Error) => {
+            this.sendError(
+              connectionContext,
+              opId,
+              { message: error.message },
+              MessageTypes.GQL_CONNECTION_ERROR,
+            );
+
+            // Close the connection with an error code, ws v2 ensures that the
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
If someone could help me workout what I broke in the tests that'd be really appreciated. This DOES fix the unhandled rejection. On my test server I do see onDisconnect called but with a `undefined` context.

OmgImAlexis:
Yeah this isn't going to work.

OmgImAlexis:
The only way I can get this to avoid the error is to avoid setting `connectionContext.initPromise = initPromise`.

Running the `initPromise` isn't the issue. It's that `connectionContext.initPromise` doesn't have a catch handler by default so if you call it with just `.then` the issue comes into play.

PR #757, opened 2020-05-03
Support persisted queries

freshollie:
Currently, although there is reference to persisted queries in the past in this repo, the client does not support the `persistedQuery` extension. When a persisted query is provided, the original query is also sent along side the `persistedQuery`, defeating the purpose.

This PR allows persistedQuery's to be transported over websocket by:

- Not passing the `query` in the payload if a persisted query is given
- Allowing a request to be made without a query
- Allowing `onOperation` on the server-side to provide the query back, as `apollo-server` should be doing the decoding.

The reasons for this are outlined here: https://github.com/apollographql/apollo-link-persisted-queries/issues/18#issuecomment-623035052

If this PR goes through, I will do the work required on `apollo-server` to give support to apollo-server for this change.

This change should not be breaking.

TODO:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### v0.9.17
+- Don't pass query in payload if persistedQuery is given
+- Allow query to be returned from `onOperation`
+- Allow query not to be given, if `persistedQuery` is passed in payload
+
 ### v0.9.16
 - Add ability to set custom WebSocket protocols for client. <br/>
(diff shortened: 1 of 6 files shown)


freshollie:
@helfer @abernix (not sure who is currently maintaining this project)

NateScarlet:
Persisted query is always if you use [apollo-link-persisted-queries](https://github.com/apollographql/):

https://github.com/apollographql/apollo-link-persisted-queries/blob/974edc3981570005d18ea455cc048f2a335c3b1b/src/index.ts#L118-L127

so full query will never been send to server.

Instead, ws link should use `operation.getContext().http.includeQuery` to decide whether to send query.

PR #775, opened 2020-06-30
fix: do not send GQL_STOP after GQL_COMPLETED is received

### LINKED ISSUE #711: GQL_STOP sent after "complete" from server
It looks like a `GQL_STOP` message is being sent from the client, _after_ receiving a completion message back from the server, resulting in the server then returning an error since the subscription is no longer active.

Screenshot 2019-12-06 at 16 08 36

My expectation would be that if the server has initiated a `complete` message, the client should silently unsubscribe.

FWIW, I'm using [gqlgen](https://gqlgen.com/) in Go for the server.

onhate:
fix #711

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### vNEXT
+- Do not send GQL_STOP when unsubscribing after GQL_COMPLETE is received [Issue #765](https://github.com/apollographql/subscriptions-transport-ws/issues/711)
+
 ### v0.9.16
 - Add ability to set custom WebSocket protocols for client. <br/>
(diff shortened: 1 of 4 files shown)


onhate:
@abernix WDYT?

onhate:
@trevorblades let me know your thoughts.

onhate:
@hwillson @stubailo could you shed a light on here? who is the responsible for this repository?

onhate:
I have publish a parallel version while that, if anyone want to get access to this fix/improvement feel free to use it

https://www.npmjs.com/package/@texelz/subscriptions-transport-ws/v/0.9.1-6.no-gql-stop

fredli74:
This change causes server memory leaks. #805

I personally do not think it's a bad client fix, but the server does not handle it correctly.

PR #809, opened 2020-09-28
fix: memory leak when executing a query or mutation

### LINKED ISSUE #805: Memory leak when executing a query or mutation
Hi,

I've run into a memory leak problem in my project, that seems to be related to the `connectionContext.operations` retaining all previous queries and mutations until the WebSocket connection is closed.

I think this problem surfaced because of the recent change in client behavior not sending GQL_STOP messages when they receive a GQL_COMPLETE message.

Looking at the server.ts implementation, it looks like all operations - not explicitly unsubscribed - will remain in memory. There seems to be no logic implemented surrounding completed operations.

onhate:
fix #805

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## vnext
+
+### Bug Fixes
+
+- Memory leak when executing a query or mutation #805
+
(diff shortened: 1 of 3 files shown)


onhate:
@hwillson could you please shed a light here, I think this is important.

